PR #14, opened 2020-06-26
feat: Prettify error output.

DanCardin:
Before, this would have been many many pages of useless internal traceback context, and (depending on the length of your migration history) a page or 2 of useless alembic stderr logs.

Screen Shot 2020-06-26 at 11 28 20 AM
Screen Shot 2020-06-26 at 11 29 57 AM


kpitzen:
This looks nice!

PR #21, opened 2021-04-21
Dc/revision data in schema

DanCardin:


kpitzen:
What's the motivation behind this, and what does it actually accomplish?

DanCardin:
the 3rd commit, which adds the ability to specify “revision data” for tables which reside in a schema. https://pytest-alembic.readthedocs.io/en/latest/custom_data.html

kpitzen (approved):

Issue #19, opened 2021-02-02
test_up_down_consistency assumes linear history and does not work with branches


adityagupta1089:

DanCardin:
As you might see in the readme for that check, I dont see the value in branched histories, although I'm sure it's something we certainly **can** accommodate in some fashion. In the interim, you can always disable that check.

Can you explain your usecase for branched history? And also, can you explain how you would navigate this check manually with alembic commands, given your history/branches?

adityagupta1089:
We have 5-6 people working on a project and each one of us creates alembic revisions specific to our changes. Usually we create a mergepoint to combine these changes.

For branched histories we can check test_up_down_consistency from each branchpoint to each connected mergepoint.

DanCardin:
As a rule, all of our histories are completely linear (which we enforce with i.e. pytest-alembic), so the existing methodology of upgrading by `+1` clearly wont work.

I assume we would have to do a kind of bfs traversal of the upgrade graph, inspecting at each node the set of available heads (or tails, going downwards :p) and upgrading each of them in turn.

I'll have to see what the workflow for that looks like in practice, given a graph that maybe branches a couple times and has nonlinear mergepoints or something.

DanCardin:
I realize this is probably too little too late, but you could easily disable (some or all of) our tests and write your own which navigates your alembic history directly, in whatever manner you'd **actually** do in real life.

I still have this in the back of my mind to solve in a more automatic way at some point, but as i said, our histories are entirely linear, so it may be a while yet.

DanCardin:
Actually, I was expecting to need to use a topological sort of the revisions, and various other things

But so I went to go write a failing test for this, and it seemed to largely revolve around the need to use "heads" instead of "head". I **think** that actually it would have previously "Just worked" if there were mergepoints already. And without mergepoints, like i said, choosing to use "heads" rather than "head" seems to have addressed it at least as far as my test examples go.

Given the age of this issue, I may just merge this if you don't respond somewhat soon. But if you do happen to notice this, ideally you'd test this against your actual history to confirm whether or not [this PR](https://github.com/schireson/pytest-alembic/pull/22) addresses your issue.

adityagupta1089:
@DanCardin pytest-alemic==0.3.0 seems to be working for my case with multiple heads in past that have been merged using mergepoints but eventually leading to only single head.

DanCardin:
[...]

PR #22, opened 2021-05-07
Fix errors caused by branched alembic history.

DanCardin:
Fixes https://github.com/schireson/pytest-alembic/issues/19


oakhan3 (approved):
👍

PR #33, opened 2021-08-25
feat: Add experimental test to identify tables which alembic will not…

### LINKED ISSUE #32: Concept: Optional built-in test which asserts all models are imported for alembic
It's trivially easy to forget to import a model such that `--autogenerate` (i.e. env.py) does not automatically recognize all the tables that *should* be included.

Perform a "clean" import of the base model MetaData (ideally through env.py itself) which collects the set of tablenames included in the MetaData. This necessarily needs to be imported in a separate "clean" python process from the current pytest execution.

Contrast that with the set of tables on the MetaData after having imported **all** child modules of some root module.

If there is any diff between the set of tables defined on either metadata, then this indicates a model/table has no import path to it from the root of the model definition.

DanCardin:
… recognize.

Closes #32 

Annoyingly, I cannot seem to get coverage to properly detect the codepaths in the subprocess call, so that's the source for essentially all of the lowered coverage


src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Thus, we start a python subprocess which does the most minimal import we can perform to extract the MetaData,
+    and return the set of tables attached to it.
+    """
+
+    command = [
+        "python",
+        "-m",
+        "pytest_alembic.tests.experimental.collect_clean_alembic_environment",
+        shlex.quote(url),
```
doctaphred, line 94: I don't think you want to quote the URL here: if you're not running with `shell=True` (which you shouldn't be, and aren't), the values of these strings show up verbatim in the subprocess' sys.argv. If this is working as-is, it's probably because none of the URLs you've passed actually contain any shell-special characters.

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+        result = subprocess.run(  # nosec
+            command,
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            env=os.environ,
+        )
+    except subprocess.CalledProcessError as e:
+        raise AlembicTestFailure(e.stderr)  # pragma: no cover
```
doctaphred, line 107: Perhaps this [anti-]pattern could be of service? https://github.com/doctaphred/phrecipes/commit/fbc07b7f8bb7a6c4dc260b2e6cc3ed546c4deeba

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+    ]
+    import os
+    try:
+        result = subprocess.run(  # nosec
+            command,
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            env=os.environ,
```
doctaphred, line 104: I think this implied: subprocesses inherit their parents' environments unless otherwise specified.

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+
+    lines = result.stdout.splitlines()
+    modules = set()
+    tablenames = set()
+
+    empty_line = 0
+    for line in lines:
+        if not line:
+            empty_line += 1
+            continue
```
doctaphred, line 117: Eww... why not just use JSON?

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
```diff
@@ -0,0 +1,191 @@
...
+    """Dynamically traverse the tree of packages below a given root and import them.
+
+    Note this will perform the import of each module, which is an operation which can
+    cause side-effects!
+    """
+
+    try:
+        package = importlib.import_module(package_name)
+    except ImportError:
+        return
```
doctaphred, line 171: These untested lines will become easily testable if you structure for DI, e.g. by adding a keyword-only `import_module=importlib.import_module` argument and passing a mock in your tests.

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
DanCardin: good to know!

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
DanCardin: yea i could, and maybe certainly should. I was caught up in trying to stay simple back when the script impl was an embedded string

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
DanCardin: they're not actually untested, I think coverage is just not sophisticated enough to follow the e2e tests. the majority of the testing of this plugin as a whole is through pytest's testdir fixture due to how the internals of alembic work which make it very challenging to actually meaningfully test alembic without on-disk histories for alembic to run against.

Although that's a fair suggestion so that i **can** get legit coverage without e2e tests, thanks!

src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py
DanCardin: ehhh, not sure it's worth the exta loc/complexity for this specific case

src/pytest_alembic/plugin/plugin.py
```diff
@@ -1,51 +1,124 @@
...
 
-    result = []
-    for test_name in test_names:
-        test_name = test_name.strip()
-        if not test_name:
-            continue
-        result.append(test_name)
-    return result
+    @classmethod
+    def collect(cls, default=True, experimental=True):
```
kpitzen, line 32: Is this `default` arg used anywhere?

DanCardin: They're both used/swapped in the helptext generation. The impl usage could probably be cleaner eventually, but it was kind of incidental to the actual feature

CHANGELOG.md
```diff
@@ -1,31 +1,81 @@
 # Changelog
 
-## v0.2.4 (2020-07-01)
+## [v0.5.0](https://github.com/schireson/pytest-alembic/compare/v0.4.0...v0.5.0) (2021-08-25)
+
+### Features
+
+* Add experimental test to identify tables which alembic will not recognize. b9fa09e
```
oakhan3, line 7:
```suggestion
* Add experimental test which will identify tables that alembic did not recognize. b9fa09e
```

CHANGELOG.md
```diff
@@ -1,31 +1,81 @@
...
+
+
+## [v0.4.0](https://github.com/schireson/pytest-alembic/compare/v0.3.3...v0.4.0) (2021-08-16)
+
+### Features
+
+* Create a mechanism in which to create multiple alembic runner fixtures. ef1d5da
```
oakhan3, line 14:
```suggestion
* Create a mechanism which allows the creation of multiple alembic runner fixtures. ef1d5da
```

README.md
```diff
@@ -75,6 +75,12 @@ itself.
   existence for a database should be able to go from an entirely blank
   schema to the finished product, and back again.
 
+- [Experimental tests
+tests](http://pytest-alembic.readthedocs.io/en/latest/experimental_tests.html)
+
+  These tests are will need to be manually enabled because their semantics or API are
```
oakhan3, line 81:
```suggestion
  These tests will need to be enabled manually because their semantics or API are
```

docs/source/experimental_tests.rst
```diff
@@ -0,0 +1,124 @@
...
+Enabling this test (TL;DR)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can either enable this test with no configuration, which will attempt to
+identify the source module from which the :code:`env.py` is loading its
+:code:`MetaData` and automatically search in that module/package
+
+.. code-block:: toml
+   :caption: pyproject.toml/setup.cfg/pytest.ini
+
+   # pyproject.tomlm
```
oakhan3, line 24:
```suggestion
   # pyproject.toml
```

docs/source/experimental_tests.rst
```diff
@@ -0,0 +1,124 @@
...
+are, regardless, imported either through this file, or through a parent package
+to this one. We'll see why later.
+
+.. code-block:: python
+   :caption: foo/bar/baz.py
+
+   from package.models import Base
+
+   class Foo/Bar/Baz(Base):
+    __tablename__ = "foo/bar/baz"
```
oakhan3, line 91: it would be easier to understand this if it was broken into multiple files and only consisted of a 2 classes/modules - foo and bar

oakhan3 (approved):

Issue #35, opened 2021-11-10
test_up_down_consistency not delivering on its promise


bluefish6:
Hi,

Thanks for creating this plugin. I've always wanted to have such tests, but never had enough time to write them in my projects.

After introducing it to my codebase, I wanted to see examples of test failures. For test `test_up_down_consistency` the description states:

>Assert that all downgrades succeed.
>While downgrading may not be lossless operation data-wise, there’s a theory of database migrations that says that the revisions in existence for a database should be able to go from an entirely blank schema to the finished product, **and back again**.

The body of the test in fact runs all upgrades and then runs all downgrades. While this confirms that the downgrades can be run successfully, it doesn't test what is the result of the downgrades. For example, such migration would not result in failure of the test (while it's clear that upgrade and downgrade are not "consistent"):

```
def upgrade():
    op.create_table(
        "foobar",
        sa.Column("id", sa.BigInteger(), nullable=False),
        sa.PrimaryKeyConstraint("id"),
    )

def downgrade():
    pass
```
The above downgrade "runs successfully", but does not deliver on the "go from an entirely blank schema to the finished product, **and back again [to an entirely blank schema]**" part. This test does not verify that the database after upgrade+downgrade is in the same state as it was before the upgrade.

**[feature request]** As a minimum it would be nice to check if, after running all downgrades, the schema is (again) empty.

If you don't want to add any logic to this test, it would be good to at least change the test description to be not misleading as to what this test checks. For example:

>Assert that all downgrades can be run successfully (without verifying that the state after upgrade+downgrade is the same as before upgrade).

Also, the test name should be changed from `test_up_down_consistency` to `test_up_down_does_not_crash`.

DanCardin:
That's definitely an good point I hadn't considered when writing the test.

In general, it's probably intractable to assert the database is the same as it was at the start (or at best, it would require a bunch of database-specific code to make this assertion. And even if we got something perfect, this test is (anecdotally) the one that's most frequently disabled, so it would probably make this test too strict to be practical anyway.

my first take for a potentially more useful approximation of the idea, would be to change the test from (essentially):
```python
 alembic_runner.migrate_up_to('head')
 alembic_runner.migrate_up_to('base')
```

to

```python
 alembic_runner.migrate_up_to('head')
 alembic_runner.migrate_up_to('base')
 alembic_runner.migrate_up_to('head')
```

[...]

PR #37, opened 2021-11-11
fix: ensure the up-down consistency test actually verifies migrations

DanCardin:
will roundtrip correctly.

Per https://github.com/schireson/pytest-alembic/issues/35, this should help but is not the whole story


tests/test_runner.py
```diff
@@ -134,3 +134,11 @@ def test_experimental_all_models_register_automatic(testdir):
     """
     result = successful_test_run(testdir, num_tests=5, test_alembic=True)
     assert "test_all_models_register_on_metadata" in result
+
+
+def test_consistency_doesnt_roundtrip(testdir):
+    """Assert a up/down consistency fails if a migration cannot rountrip up -> down -> up."""
```
luke-mino-altherr, line 140: Would we want to go up->down->up->down to leave the db in the down state?

DanCardin: It shouldn't really matter, i dont think (unless i'm missing your point.

The test itself is verifying that your downgrade is putting the database in a similar enough state that your up works the 2nd time (which would generally catch typical up migrations with DDL changes).

But if you're just talking about cleanup-wise; in general in these tests, we leave the database in whatever state the test happens to leave it in. It'd be the job of the fixture infrastructure to either clean up (or in pytest-mock-resource's case, hand you a clean database the next time around).

luke-mino-altherr (approved):

Issue #44, opened 2021-12-17
asyncio support?


adaamz:
Hi,
I would like to ask if is it possible to use asynchronous sqlalchemy with pytest-alembic. If yes then how to configure it, because I'm not able to configure it proper way.

This is how I create session in tests:
```python
@pytest.fixture()
async def async_session_empty(config):
    engine = create_async_engine(config.db.url)
    async_sess = sessionmaker(engine, expire_on_commit=False, class_=AsyncSession)
    async with async_sess() as session:
        yield session

    await engine.dispose()
```

I tried this
```python
@pytest.fixture()
@pytest.mark.asyncio
async def alembic_engine(async_session_empty):
    return async_session_empty


@pytest.fixture()
def alembic_config(config):
    # use this in env.py
    return {"script_location": "alembic", "db_url": config.db.url}
```
but I only get some pytest warning about unclosed socket
```
ResourceWarning: unclosed <socket.socket fd=22, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('172.20.0.6', 57792), raddr=('172.20.0.3', 5432)>
```

Thank you for your response.

DanCardin:
The answer is going to minimally include applying https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html to your `env.py` if you haven't already.

Although in my attempt at writing a test-case for this scenario so far, I am personally seeing:
```
tests::pytest_alembic::test_model_definitions_match_ddl <- . FAILED      [ 25%]
tests::pytest_alembic::test_single_head_revision <- . PASSED             [ 50%]
tests::pytest_alembic::test_up_down_consistency <- . FAILED              [ 75%]
tests::pytest_alembic::test_upgrade <- . FAILED                          [100%]

=================================== FAILURES ===================================
___ [pytest-alembic] tests::pytest_alembic::test_model_definitions_match_ddl ___
Alembic Error:
    Task <Task pending name='Task-3' coro=<run_migrations_online() running at migrations/env.py:27> cb=[_run_until_complete_cb() at ...python3-3.8.8/lib/python3.8/asyncio/base_events.py:184]> got Future <Future pending cb=[Protocol._on_waiter_completed()]> attached to a different loop
```
So it appears to be slightly working.

In any case, the internals of any of the fanciers features of this plugin are currently assuming sync sqlalchemy, so once I get a basic test passing, i'll have to noodle around on whether the rest is supportable

[...]

PR #45, opened 2021-12-20
test: Support asyncio based test execution.

DanCardin:
Adds examples/tests for async usage of alembic.

A few changes needed to be applied to make this compatible with ad-hoc tests and test_downgrade_leaves_no_trace, but all 4 vanilla tests already worked out of the box.

The examples (the vast majority of the file changes) are largely copy-pasted from prior examples, and aren't particularly interesting to read.

Fixes: https://github.com/schireson/pytest-alembic/issues/44

* BREAKING CHANGE: Starting with this release, python 3.6 will no longer be tested or officially supported. In this specific release, only the new official support for asyncio-based engine with alembic and pytest-alembic is incompatible with 3.6. Any existing usage should remain at least provisionally compatible until later releases which may or may not further break compatibility.

### DIFF
--- a/docs/source/asyncio.rst
+++ b/docs/source/asyncio.rst
@@ -1,0 +1,134 @@
+Asyncio
+=======
+
+Support for asyncio is largely built on top of the `Alembic Cookbook`_ example,
+inlined here for posterity:
+
+.. code-block:: python
+
+   import asyncio
+
+   # ... no change required to the rest of the code
+   
+   def do_run_migrations(connection):
+       context.configure(connection=connection, target_metadata=target_metadata)
+   
+       with context.begin_transaction():
+           context.run_migrations()
+   
+   
+   async def run_migrations_online():
+       """Run migrations in 'online' mode.
+   
+       In this scenario we need to create an Engine
+       and associate a connection with the context.
+   
+       """
+       connectable = AsyncEngine(
+           engine_from_config(
+               config.get_section(config.config_ini_section),
+               prefix="sqlalchemy.",
+               poolclass=pool.NullPool,
+               future=True,
+           )
+       )
+   
+       async with connectable.connect() as connection:
+           await connection.run_sync(do_run_migrations)
+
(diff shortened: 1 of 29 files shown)


src/pytest_alembic/executor.py
```diff
@@ -50,21 +49,28 @@ def run_command(self, command, *args, **kwargs):
         return self.stdout.readlines()
 
 
-@dataclass(frozen=True)
+@dataclass
 class ConnectionExecutor:
-    connection: Connection
+    metadatas: Dict[str, MetaData] = field(default_factory=dict)
```
DanCardin, line 55: can no longer cache/store the connection because it can change call-to-call on the same revision.

docs/source/asyncio.rst
```diff
@@ -0,0 +1,134 @@
...
+Note that this is a prerequisite for how one gets **alembic itself** to run with an async
+connection, when running ``alembic`` commands interatively yourself.
+
+At this point, you just need to make sure the ``alembic_engine`` fixture is producing a async engine.
+something like
+
+.. code-block:: python
+
+   from sqlalchemy import create_engine
+   from sqlalchemy.ext.asyncio import create_engine_async, AsyncEngine
```
kpitzen, line 56: Do we want to include this `AsyncEngine` import in the above example as well?  At least for completeness.

kpitzen (approved):

docs/source/asyncio.rst
DanCardin: I copied that example directly from alembic's docs, so perhaps not, so i dont have to maintain a diff, if it ever changes (and it's going [to change](https://github.com/sqlalchemy/alembic/pull/973))

CaselIT: we can update it in the alembic docs too.

CaselIT: added here https://github.com/sqlalchemy/alembic/commit/1364bbf9b782178cad66ca22783dde1ba8353fde

PR #47, opened 2022-01-21
docs: update example for insert_to

KyleKing:
Thanks for building and maintaining this plugin! It is really useful

I found a small error in the documentation on how to use `insert_to` because the `tablename` kwarg appears to be removed:

https://github.com/schireson/pytest-alembic/blob/fac82f1742503de921d4d559bb05f99112e430e7/examples/test_migrate_up_before/test_migrations.py#L4


DanCardin (approved):

DanCardin:
thanks for the kind words and thanks for the contribution!

PR #73, opened 2023-02-21
 fix: Add an option to configure the default test registration path. 

DanCardin:


oakhan3 (approved):
it would be possible better API if `--test-alembic` took in an optional path imo, but this works just as well

DanCardin:
I dont necessarily disagree, i just dont think i can backwards compatibly apply that, and/or whether pytest's option system lets you do that.

PR #112, opened 2024-07-29
Updated pyproject.toml to be more flexible with poetry_core versioning

imadnyc:
This dependency works with bigger versions, and is convenient to nix users since the version in [nixpkgs is the latest core.](https://search.nixos.org/packages?channel=24.05&show=python312Packages.poetry-core&from=0&size=50&sort=relevance&type=packages&query=poetry+core) I think having compatibility with newer versions is generally useful too.


DanCardin (approved):

DanCardin:
Fair enough, thanks!

PR #123, opened 2025-04-01
Take snapshot before any changes are made

JarrodBaker:
Assume we're starting at `M_0`. In the existing implementation `check_revision_cycle` applies the next migration `M_1`, takes a snapshot and then rolls it back. So our _initial_ view of the database was after the migration had been applied with the `head` at `M_1`.

Then it would go up (`M_1`) and down once (`M_0`) before taking a new snapshot. The second snapshot is at `M_0`. 

The two snapshots are at `M_1` and `M_0` respectively. These were compared but they were never the same.
 
I tested this with a very simple migration adding a single table and a single column. The test fails out of the box.


DanCardin:
Do you have an example migrations history that exemplifies the problem? clearly the existing success/failure test is not rigorous enough if your PR and main are both passing tests.

i believe the point **is** to take the snapshot post-upgrade. apply the upgrade, snapshot it, then use autogenerate to compare current (pre migration) state to dest (post upgrade) state. the resultant autogenerate should resemble the upgrade you actually applied (but we can't know whether the user's upgrade has alterations).

i think your PR is effectively diffing current state vs the input metadata to produce the initial diff. whereas `main` is diffing current state vs the actual applied migration state's metadata (which could be different).

JarrodBaker:
I tried to create a minimalist example and the existing code passes on the same migration that failed within the internal project that I was working on. In our existing code-base the problem was any migration, even as simple as adding and removing a table with one column.

The problem seems to be I was missing a step from [the setup instructions](https://pytest-alembic.readthedocs.io/en/stable/setup.html#env-py) checking the the `connectable` is `None`. Once I added that in things worked as expected. 

Thanks also for your explanation of the intent of the logic of the test in the comment above: I have a better understanding of what it is the way it is now.

PR #128, opened 2025-05-14
Revert test commits

mschoettle:
Revert test commits from #126.

Dropped `pytest 6`, `sqlalchemy 1.3.0`, and Python `3.7-3.8`. All due to incompatibilities (potentially also due to upgrading poetry in #126).


mschoettle:
@DanCardin

DanCardin:
i would ideally like to support sqlalchemy 1.3 and 3.7/8 indefinitely because they really dont pose all that much of a maintenance burden, but so much of the ecosystem is starting to require it that it's realistically not that big a deal. Seems fine.

DanCardin:
although, having just merged it, did you mean to remove 3.13 again?

mschoettle:
I had trouble figuring out why the jobs failed due to too many changes. So I split them. 3.13 is in #129.

PR #129, opened 2025-05-14
Add support for Python 3.13

### LINKED ISSUE #127: Support Python 3.13
When running `pytest-alembic` with Python 3.13 the following exception is raised:

```
>                   warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
E                   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Connection object at 0x7fb50ac162f0>
E                   
E                   Traceback (most recent call last):
E                     File "<string>", line 5, in __init__
E                   ResourceWarning: unclosed database in <sqlite3.Connection object at 0x7fb50ac162f0>
```

See a full pytest run here: https://github.com/mschoettle/pytest-alembic/actions/runs/15021736015/job/42212325116

mschoettle:
SQLite connection needs to be explicitly closed: https://docs.python.org/3/library/sqlite3.html#how-to-use-the-connection-context-manager

Fixes #127

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         pytest-version: ["7.0.0", "8.3.5"]
         pytest-asyncio-version: ["0.16.0", "0.23.0"]
-        sqlalchemy-version: ["1.4.0", "2.0.40"]
+        sqlalchemy-version: ["2.0.40"]
 
         exclude:
           # old versions of pytest-asyncio use old pytest hook syntax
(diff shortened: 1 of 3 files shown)


mschoettle:
@DanCardin 1.4.0 does not seem to have `Connection.commit`. We could of course also drop it and require SQLAlchemy 2.x.

I saw your other comment about supporting indefinitely: Those that want to use older versions could stick to the current version of `pytest-alembic`. Those who use current versions can use the latest version(s).

mschoettle:
Also, any idea why it is still complaining about the unclosed database?

DanCardin:
It was the default fixture engine. It was failing at the end of the test run each time, rather than on a specific test, which led me to the discovery.

mschoettle:
Nice! Thanks for the help

mschoettle:
@DanCardin Is there any plan to cut a new release with these changes?

DanCardin:
done!

Issue #131, opened 2025-05-22
pytest-alembic can't handle alembic settings in pyproject.toml


hmvp:
See https://alembic.sqlalchemy.org/en/latest/changelog.html#change-96833a0fc8de30bdd78c897205b5c77d for the new option

When using pyproject.toml instead of alembic.ini pytest-alembic does not find the correct settings

Using a config fixture as workaround also does not work:

```python
@pytest.fixture
def alembic_config():
    return Config(toml_file="pyproject.toml")
```

This is  because `def make_alembic_config(self, stdout):` uses `alembic_config.get_section_option("alembic", "script_location")` and not `alembic_config.get_alembic_option("script_location")` (which is new) to get the script location...

If both the fixture is set and the above line is changed is all works...

DanCardin:
Any chance you could sanity check that this solves the problem before I merge/release?

hmvp:
Sure, but that wil be tomorrow for me...

hmvp:
Thanks for the package and the quick response!!! Much appreciated...

PR #132, opened 2025-05-22
fix: Handle pyproject.toml based alembic config.

DanCardin:
Fixes https://github.com/schireson/pytest-alembic/issues/131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,18 @@
 # Changelog
 
-## [Unreleased](https://github.com/schireson/pytest-alembic/compare/v0.11.0...HEAD) (2025-05-16)
+### [v0.12.1](https://github.com/schireson/pytest-alembic/compare/v0.12.0...v0.12.1) (2025-05-22)
+
+#### Fixes
+
+* Handle pyproject.toml based alembic config.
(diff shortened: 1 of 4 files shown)


hmvp:
Seems to work for me, both with old version of alembic and settings ini, new version and settings in ini and new version and settings in pyproject.toml

The only thing is that you do need a alembic_config fixture in that case to point to pyproject.toml as the source of the config otherwise it is not found.

When I change the start of  `make_alembic_config` to:

```python
    def make_alembic_config(self, stdout):
        ini_file = (
            self.config_options.get("file")
            or self.config_options.get("config_file_name")
            or "alembic.ini"
        )

        alembic_config = self.config_options.get("alembic_config")

        if not alembic_config and self.alembic_config:
            alembic_config = self.alembic_config
            alembic_config.stdout = stdout
        else:
            if "toml_file" in inspect.getfullargspec(alembic.config.Config.__init__).args:
                alembic_config = alembic.config.Config(file_=ini_file, toml_file="pyproject.toml", stdout=stdout)
            else:
                alembic_config = alembic.config.Config(file_=ini_file, stdout=stdout)

```

That is not needed... 

However with al the possible config options I am not sure if that does not break for specific cases...

hmvp:
Looks good to me, all situations tested now work for me!

PR #143, opened 2026-03-02
fix: Support anyio.

### LINKED ISSUE #119: Incompatible with anyio due to event loop keep-alive for session fixtures
`anyio` creates an event loop during the fixture, and that loop will be reused/kept alive until the fixture is cleaned up (see https://anyio.readthedocs.io/en/stable/testing.html#technical-details). This means that if you have a session scooped fixture (which is quite common, as you might have a DB setup fixture):

```python
@pytest.fixture(scope="session", name="db")
async def _setup_db() -> None:
    if await database_exists(ENGINE.url):  # pragma: no branch
        await drop_database(ENGINE.url)  # pragma: no cover
    await setup_db()
```

Then this fixture will keep alive the event loop until the end of the session (such fixture requires all your tests using the DB to reuse the same event loop).

`pytest-alembic`, however, assumes that when its test is running, there is no other event loop running, as it is calling https://github.com/schireson/pytest-alembic/blob/5ff6ca08f553925095389da6048895d41de2b505/src/pytest_alembic/executor.py#L182 which is only allowed to be called when there is no other live event loop. 

Instead, `pytest-alembic` IMHO should check if there is a live event loop and reuse it if so.

DanCardin:
Maybe fix #119

### DIFF
--- a/examples/test_async_anyio/alembic.ini
+++ b/examples/test_async_anyio/alembic.ini
@@ -1,0 +1,37 @@
+
+[alembic]
+script_location = migrations
+
+[loggers]
+keys = root,sqlalchemy,alembic
+
+[handlers]
+keys = console
+
+[formatters]
+keys = generic
+
+[logger_root]
+level = WARN
+handlers = console
+qualname =
+
+[logger_sqlalchemy]
+level = WARN
+handlers =
+qualname = sqlalchemy.engine
+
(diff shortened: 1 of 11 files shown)


tschm:
@DanCardin please close this one. It contains poetry.lock and serious work would be required to update it

PR #148, opened 2026-08-15
feat: add type hints to exposed tests

adhtruong:
This PR adds type coverage to tests that are part of the public interface. This avoids issues in usage when strict mode is enabled in pyright and other tools


DanCardin (approved):

DanCardin:
appreciate it!

DanCardin:
CI was failing due to not having a recent build, so i converted to uv since i'd wanted to anyways, and rebased you

PR #184, opened 2026-08-25
chore!: drop Python 3.9, raising the floor to 3.10

### LINKED ISSUE #183: Drop Python 3.9 support (EOL since 2025-10-31), raising the floor to 3.10
**Subcategory:** public API / semver discipline, toolchain currency

## Why

Python 3.9 reached end of life on **2025-10-31** (PEP 596) — roughly ten months ago. It
receives no further security patches, and `requires-python = ">=3.9,<4"`
(`pyproject.toml:10`) is now the only thing holding the project's language floor three
releases below the oldest supported Python.

The cost is not hypothetical. It is currently pinning the lint configuration to a
version that predates the floor by *three years*:

```toml
[tool.ruff]
target-version = "py37"   # pyproject.toml:52
```

`"UP"` (pyupgrade) is selected under `[tool.ruff.lint]`, but with `target-version` at
`py37` those rules cannot suggest anything newer than 3.7 syntax. So the modernisation
rules the project has opted into are silently inert.

## What dropping 3.9 unlocks

**1. A truthful `target-version`.** `py310` lets the already-selected `UP` rules
actually fire.

**2. PEP 604 / PEP 585 typing throughout.** Eight modules still import the deprecated
typing aliases:

```
src/pytest_alembic/config.py:2      Any, cast, Dict, List, Optional, TYPE_CHECKING, Union
src/pytest_alembic/executor.py:5    Dict, List, Optional, Union
src/pytest_alembic/history.py:4     Dict, List, Optional, Tuple
src/pytest_alembic/revision_data.py:2   Dict, List, TYPE_CHECKING, Union
src/pytest_alembic/plugin/plugin.py:12  Callable, cast, Dict, List, Optional
src/pytest_alembic/plugin/fixtures.py:1 Any, Dict, Union
src/pytest_alembic/plugin/error.py:4    List
src/pytest_alembic/tests/experimental/all_models_register_on_metadata.py:7  List, Optional, Set, Tuple
```

`runner.py` already writes the modern form (`table: str | None`, `dict | list | None` at
`runner.py:274`, `:318`, `:366`) but needs `from __future__ import annotations` to do it
on 3.9. On a 3.10 floor that import becomes unnecessary and the rest of the package can
match, rather than the codebase carrying two typing dialects.

**3. A smaller CI matrix.** 15 entries today, 12 without 3.9 — a 20% reduction in jobs
per push, measured from `build.yml`:

| | entries |
| --- | ---: |
| current | 15 |
| without 3.9 | 12 |

## What to change

- `pyproject.toml:10` — `requires-python = ">=3.10,<4"`
- `pyproject.toml:52` — `target-version = "py310"`, then run `make format` and let the
  `UP` rules rewrite the typing imports
- `.github/workflows/build.yml` — drop `"3.9"` from `matrix.python-version`
- `docs/source/contributing.rst:15-16` — states "supports Python 3.9 and above" and "CI
  exercises 3.9 through 3.13"; both become wrong

There is no version-conditional code to remove: `grep -rn "version_info" src/` returns
nothing, so the only 3.9 accommodations are the typing dialect and the `__future__`
import.

## Adjacent, and worth doing at the same time

- `types-dataclasses>=0.1.7,<0.7` (`pyproject.toml:36`) is a stub package for the
  `dataclasses` **backport**, which has been stdlib since 3.7. This is already dead
  weight today — not something dropping 3.9 unlocks — but it is the same edit.
- `[project]` declares only `classifiers = ["Framework :: Pytest"]`, with no
  `Programming Language :: Python :: 3.x` entries at all, so nothing there needs
  correcting — though adding them would make the supported range visible on PyPI.

## Compatibility note

Raising `requires-python` is a breaking change for anyone still on 3.9 and belongs in a
version bump rather than a patch. It is a *safe* break, though: pip and uv read
`requires-python` and refuse to install the new release on 3.9, so existing users are
pinned to the last compatible version rather than getting a broken install.

## done when

`requires-python` is `>=3.10`, `target-version` is `py310`, the CI matrix no longer
contains 3.9, `contributing.rst` states the new floor, and `make lint` and `make test`
pass with the typing imports modernised.

tschm:
Closes #183.

Python 3.9 has been EOL since 2025-10-31 (PEP 596). `requires-python = ">=3.9,<4"` was the only thing holding the floor three releases below the oldest supported Python, and it kept `[tool.ruff] target-version` at `py37` — so the `UP` (pyupgrade) rules the project selects could not suggest anything newer than 3.7 syntax and were silently inert.

## What changed

- **`pyproject.toml`** — `requires-python = ">=3.10,<4"`, `target-version = "py310"`. `types-dataclasses` dropped from the dev group (it stubs the `dataclasses` *backport*, stdlib since 3.7). `classifiers` gains `Programming Language :: Python :: 3.10`–`3.13` so the supported range shows on PyPI.
- **Typing dialect** — with `UP` live, the eight modules importing `Dict`/`List`/`Optional`/`Union` move to PEP 585 builtin generics and PEP 604 unions. `runner.py` drops `from __future__ import annotations`; its `Config` forward references are quoted instead, which is the convention `config.py` and `revision_data.py` already use. The package now carries one dialect rather than two.
- **`history.py`** — `revision_window` uses `itertools.pairwise` (3.10+) in place of the hand-rolled `tee`/`islice`/`zip`, which also resolves the `B905` (`zip()` without `strict=`) that `py310` newly reports. `import collections` becomes unused and goes.
- **`.github/workflows/build.yml`** — `"3.9"` out of the matrix: 15 entries to 12.
- **`docs/source/contributing.rst`** — states the 3.10 floor.
- **`uv.lock`** — relocked; the 3.9 resolution markers and their duplicate dependency entries drop out.

## Verification

`make lint` (ruff check, ruff format --check, mypy), `make test` (122 passed, 95% coverage) and `make docs-coverage` (71.7%, floor 70%) all pass locally on 3.13.

## Not included

Raising `requires-python` is a breaking change and belongs in a minor bump, but `version` in `pyproject.toml` is edited by hand and releases are tag-driven, so the bump from `0.12.1` is left as a separate release decision. It is a *safe* break either way: pip and uv read `requires-python` and refuse the new release on 3.9, so 3.9 users stay pinned to the last compatible version rather than getting a broken install.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         pytest-version: ["7.0.0", "8.3.5"]
         pytest-asyncio-version: ["0.16.0", "0.23.0"]
         sqlalchemy-version: ["2.0.40"]
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -12,8 +12,8 @@ you will also need Docker running locally.
 
 Getting Setup
 -------------
-``pytest-alembic`` supports Python 3.9 and above (see ``requires-python`` in
-``pyproject.toml``). CI exercises 3.9 through 3.13 against a matrix of ``pytest``,
+``pytest-alembic`` supports Python 3.10 and above (see ``requires-python`` in
+``pyproject.toml``). CI exercises 3.10 through 3.13 against a matrix of ``pytest``,
 ``pytest-asyncio`` and ``sqlalchemy`` versions.
 
 Run :code:`make help` to list the common commands, but for some basic setup:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,9 +5,15 @@ description = "A pytest plugin for verifying alembic migrations."
 authors = [{ name = "Dan Cardin", email = "ddcardin@gmail.com" }]
 license = { text = "MIT" }
 keywords = ["pytest", "sqlalchemy", "alembic", "migration", "revision"]
-classifiers = ["Framework :: Pytest"]
+classifiers = [
+    "Framework :: Pytest",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+]
 readme = "README.md"
-requires-python = ">=3.9,<4"
+requires-python = ">=3.10,<4"
 dependencies = ["pytest>=7.0", "alembic", "sqlalchemy"]
 
 [project.urls]
@@ -33,7 +39,6 @@ dev = [
     "pytest-mock-resources[docker]>=2.6.3",
     "ruff>=0.15.0",
     "sqlalchemy[asyncio]>=2.0.42",
-    "types-dataclasses>=0.1.7,<0.7",
     "anyio",
 ]
 docs = ["myst-parser", "sphinx", "sphinx-rtd-theme", "sphinx-autodoc-typehints", "sphinx-autobuild"]
@@ -49,7 +54,7 @@ exclude_lines = ["pragma: no cover", "if TYPE_CHECKING:", "if __name__ == .__mai
 
 [tool.ruff]
 src = ["src", "tests"]
-target-version = "py37"
+target-version = "py310"
 line-length = 100
 
 [tool.ruff.lint]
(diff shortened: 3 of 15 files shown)


tschm:
@DanCardin Please be careful here. The drop of 3.9 should be done. 3.10 can be supported for the moment. There is an open issue for supporting 3.14

DanCardin:
I left the syntax to 3.7 at the time as a way of provisionally allowing it to work in 3.7 without needing to test it. And ditto haven't found supporting 3.9 to be too onerous a task which is why i haven't bumped it till now. with that said https://pypistats.org/packages/pytest-alembic does show **some** 3.9 usage, probably most of it is CI for this repo tbh.

PR #185, opened 2026-08-25
docs: document every symbol, taking docstring coverage to 100%

tschm:
The interrogate ratchet stood at 70% against a measured 71.7% — 34 of 120 symbols undocumented. This documents all 34 and moves `fail-under` to 100, so the floor *is* the measured number and cannot erode silently.

## What was written

**Eight module docstrings**, each saying what the module is for and how it relates to its neighbours: the two executors, the flattened history, the revision data, the configuration surface, the fixtures, the default tests, and the two experimental tests.

**Twenty-six functions and classes.** The ones carrying behaviour that isn't obvious from the signature got the most attention:

| Symbol | What the docstring records |
| --- | --- |
| `CommandExecutor.run_command` | Why stderr is discarded, and why the buffer position is recorded before the command rather than read from the start |
| `ConnectionExecutor.table` / `table_insert` | The per-revision reflection cache, and the `__tablename__` key that overrides the `tablename` argument (and may be schema-qualified) |
| `AlembicHistory.validate_revision` + the four walk methods | `head` coercing to `heads`, and what `None` means at each end of the range |
| `check_revision_cycle` | Why the comparison is between two *rendered DDL diffs* rather than two reflected `MetaData` objects, and why every step rolls back even on success |
| `WrappingConnection` | That alembic's `env.py` calls `connect()`, which would otherwise open a second connection outside the transaction the test is holding open |
| `Config.make_alembic_config` | The precedence order: `config_options` wins over whatever the ini/toml file said |

Plus `duplicate_alembic_config`, `parse_collection_output`, `url_to_string`, `create_connectable`, the two `from_config` constructors, `set_revision`, `stamp`, `heads`, `configure`, `execute_fn` and `RevisionData.get`.

## Two new doctests

`revision_window` and `set_revision`. `pytest` runs docstrings under `--doctest-modules`, so these execute rather than decorate — test count goes 122 → 124.

## The floor

`fail-under` goes 70 → 100, and the comment above it now explains why a 100% floor is the one that can't silently erode. The exclusions below it are unchanged, so meeting it doesn't mean writing docstrings for `__repr__`: `ignore-magic`, `ignore-private`, `ignore-semiprivate`, `ignore-nested-functions` and `ignore-init-module` still apply, and `D1` stays ignored under `[tool.ruff.lint]` so no individual symbol is *required* to carry one — interrogate gates the proportion, ruff gates the style.

## Verification

`make lint` (ruff check, ruff format --check, mypy) clean, `make test` 124 passed at 95% line coverage, `make docs-coverage` reports 120/120 against the new 100% floor.

No behaviour changes — docstrings and one config value.


DanCardin (approved):

tschm:
@DanCardin I have fixed the conflicts...

PR #194, opened 2026-08-25
test: take coverage to 100% and raise the floor to match

tschm:
**Builds on #193 — merge that first.** This branch contains #193's commit as its parent, so the diff below currently shows both; once #193 merges, this PR's diff collapses to its own single commit automatically. (A stacked base is not possible here: the base of a cross-fork PR has to be a branch in this repository, and only `main` is.) It is stacked rather than independent because both changes touch nearly every line of `tests/`, and new tests have to be annotated to satisfy #193's `disallow_untyped_defs` anyway.

**Review the second commit**, `test: take coverage to 100% and raise the floor to match`.

## Why

`fail_under = 93` against ~95% measured left 35 unexecuted statements and 15 partial branches with nowhere to show up. That is the same slack `[tool.interrogate]` was ratcheted out of in #185 — a floor below the measured number lets a new untested path hide inside the gap instead of failing the build.

The comment on the old floor was also stale:

> Held below the ~95% actually observed locally, because the CI matrix skips the asyncio tests on sqlalchemy versions without asyncio support

The matrix now pins a single sqlalchemy (`2.0.40`), so no entry skips those tests any more.

## Most of the gap was reachable, and is now tested

24 new tests, in three new files and five existing ones:

| what | where |
| --- | --- |
| the pre-`pyproject.toml` alembic path, both branches of `_supports_toml` | `test_config.py` |
| `CommandError` → `RuntimeError`; a row naming no table | `test_executor.py` |
| `raw_command`; `insert_into` with no data; `generate_revision` not refreshing history; both halves of `managed_downgrade`'s error handling | `test_migration_context.py` *(new)* |
| the disabled-plugin session hook, and the enabled one | `plugin/test_hooks.py` *(new)* |
| chained `include`/`include_experimental`/`exclude`; an item with no fixtures | `plugin/test_plugin.py` |
| `parse_collection_output` with no payload; `url_to_string`'s two fallbacks | `test_all_models_register_on_metadata.py` |
| `create_connectable`, sync and async | `test_collect_clean_alembic_environment.py` |
| `WrappingConnection` | `test_downgrade_leaves_no_trace.py` *(new)* |

`test_migration_context.py` is stub-driven on purpose, and says so: `tests/test_runner.py` drives real histories through `pytester`, which is right for anything where migrations actually run, but these paths need a particular *executor response* rather than a particular history.

## One test was silently doing nothing

`test_revision_spec_input` passed its specs inside `config_options` — the first positional argument of `Config` — so they never reached `before_revision_data`/`at_revision_data`, `RevisionSpec.parse` was never called with a `RevisionSpec`, and the test asserted nothing at all. It now constructs the fields it meant to and asserts that `parse` hands back the very object it was given.

## Two dead branches removed rather than tested

Both were unreachable, so a test for them could not be written honestly:

- `MigrationContext.current` ended `if current: return current / return "base"`. `current` is initialised to `"base"` and `get_current` only ever reassigns it to a revision hash, so the guard can never be False.
- `test_model_definitions_match_ddl` re-tested `migration_is_empty` inside the block that had just established it.

## Two exclusions, with the reasoning recorded at the site

- **The version-dispatched `pytest_collect_file` variants.** Exactly one of the three is defined in any given interpreter, so no single run can execute the other two — and since the matrix runs both pytest 7 and pytest 8 and *each entry enforces the floor on its own*, a run that covers all three does not exist. What they do is exercised through the pytester-based tests either way.
- **The asyncio `ImportError` fallback**, matching the `DeclarativeMeta` one immediately above it, which was already excluded for the same reason.

## Verification

- `make test` → **148 passed**, `TOTAL 895 0 208 0 100%` — every statement and every branch, in every file.
- `make lint` → clean on 34 files, including the new tests under #193's `disallow_untyped_defs`.
- `make docs-coverage` → 100%, unchanged.

## One thing to know about the floor

Because each matrix entry enforces it independently, anything version-conditional has to be excluded for *every* entry, not just this one. The remaining sensitivity is `test_pyproject_toml`, which skips itself when alembic lacks `get_alembic_option` — alembic is unpinned in the matrix so it always resolves modern, but if it ever did not, that entry would drop below 100 rather than merely losing a test.

### DIFF
--- a/src/pytest_alembic/tests/default.py
+++ b/src/pytest_alembic/tests/default.py
@@ -82,22 +82,21 @@ def test_model_definitions_match_ddl(alembic_runner: MigrationContext) -> None:
             autogen_context = AutogenContext(migration_context)
             rendered_upgrade = _render_cmd_body(script.upgrade_ops, autogen_context)
 
-            if not migration_is_empty:
-                message = (
-                    "The models describing the DDL of your database are out of sync with the set of "
-                    "steps described in the revision history. This usually means that someone has "
-                    "made manual changes to the database's DDL, or some model has been changed "
-                    "without also generating a migration to describe that change."
-                )
-                raise AlembicTestFailure(
-                    message,
-                    context=[
-                        (
-                            "The upgrade which would have been generated would look like",
-                            rendered_upgrade,
-                        )
-                    ],
-                )
+            message = (
+                "The models describing the DDL of your database are out of sync with the set of "
+                "steps described in the revision history. This usually means that someone has "
+                "made manual changes to the database's DDL, or some model has been changed "
+                "without also generating a migration to describe that change."
+            )
+            raise AlembicTestFailure(
+                message,
+                context=[
+                    (
+                        "The upgrade which would have been generated would look like",
+                        rendered_upgrade,
+                    )
+                ],
+            )
 
     test_upgrade(alembic_runner)
     alembic_runner.generate_revision(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,10 +46,19 @@ docs = ["myst-parser", "sphinx", "sphinx-rtd-theme", "sphinx-autodoc-typehints",
 [tool.coverage.report]
 show_missing = true
 skip_covered = true
-# Ratcheted floor. Held below the ~95% actually observed locally, because the CI
-# matrix skips the asyncio tests on sqlalchemy versions without asyncio support,
-# which legitimately lowers coverage on those entries.
-fail_under = 93
+# The ratchet has reached the top, like [tool.interrogate] below: every statement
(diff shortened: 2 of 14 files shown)


src/pytest_alembic/tests/default.py
```diff
@@ -82,22 +82,21 @@ def verify_is_empty_revision(migration_context: Any, __: Any, directives: Any) -
...
-                raise AlembicTestFailure(
-                    message,
-                    context=[
-                        (
-                            "The upgrade which would have been generated would look like",
-                            rendered_upgrade,
-                        )
-                    ],
-                )
+            message = (
```
DanCardin, line 85: wat was i doing

DanCardin (approved):

PR #215, opened 2026-08-26
chore: Revise testing matrix.

### LINKED ISSUE #176: Extend the CI matrix to Python 3.14 (needs a pytest 9 / pytest-asyncio 1.x axis)
**Subcategory:** toolchain reproducibility / CI coverage
**Follow-up to:** #168 (see #175)

## Why

#175 pinned `.python-version` to 3.13, closing the local-vs-CI divergence. It deliberately
did **not** extend the matrix to Python 3.14, because that is not a one-line change. This
issue tracks the remaining half: CI does not test the newest interpreter the package
actually supports.

The package *does* work on 3.14 — the entire #164/#165/#169 work was developed and verified
on 3.14.6 with pytest 9.1.1, full suite green. So this is a coverage gap, not a support
gap, which is why `requires-python` was not narrowed to `<3.14`.

## What breaks, and why it is not one line

Both failure modes were confirmed empirically in throwaway 3.14 environments, so the next
person does not have to rediscover them.

**1. pytest 7.0.0 cannot run on 3.14 at all.** Its assertion rewriter calls `ast.Str`,
removed from the stdlib:

```
_pytest/assertion/rewrite.py:943: in visit_Assert
    assertmsg = ast.Str("")
E   AttributeError: module 'ast' has no attribute 'Str'
```

Installation succeeds (`pytest==7.0.1`), so this only surfaces at collection time.

**2. pytest-asyncio 0.23.x fails under this project's `filterwarnings = error`.** With
pytest 8.3.5 + pytest-asyncio 0.23.8 on 3.14:

```
E   pluggy.PluggyTeardownRaisedWarning: A plugin raised an exception during an
    old-style hookwrapper teardown.
E   Plugin: asyncio, Hook: pytest_pycollect_makeitem
E   DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for
    removal in Python 3.16; use inspect.iscoroutinefunction() instead
```

A second one fires from `asyncio.get_event_loop_policy`. Importantly, the existing filter
`ignore:.*is deprecated and will be removed in Python 3.14.*:DeprecationWarning`
(`pyproject.toml`) does **not** match these, because the messages say *"slated for removal
in Python 3.16"*.

So of the three pytest/asyncio combinations the matrix currently runs per interpreter,
**two break on 3.14**. Only pytest 9 + pytest-asyncio 1.x works — which is exactly the
combination `uv sync` resolves locally, and exactly the one CI never exercises.

## Suggested shape

Add a pytest 9 / pytest-asyncio 1.x axis and exclude the combinations proven broken:

```yaml
python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
pytest-version: ["7.0.0", "8.3.5", "9.1.1"]
pytest-asyncio-version: ["0.16.0", "0.23.0", "1.4.0"]

exclude:
  # old versions of pytest-asyncio use old pytest hook syntax
  - pytest-version: "8.3.5"
    pytest-asyncio-version: "0.16.0"
  # pytest 7's assertion rewriter uses ast.Str, removed in the stdlib by 3.14
  - python-version: "3.14"
    pytest-version: "7.0.0"
  # pytest-asyncio 0.23.x hits deprecated asyncio APIs that `filterwarnings = error`
  # escalates on 3.14
  - python-version: "3.14"
    pytest-asyncio-version: "0.23.0"
```

The exact exclude set needs iterating against real CI runs — the combinations above were
verified for 3.14 but the new pytest 9 axis crossed with 3.9–3.13 is unverified, and
pytest 9 may well need its own excludes against older interpreters.

Worth confirming the job count stays reasonable while doing so; the matrix is currently 15
jobs (5 interpreters × 3 surviving combinations) and a naive 6 × 9 expansion would be far
larger.

**done when…** CI runs at least one green 3.14 job, the excludes carry comments naming the
incompatibility each one works around, and the job count is deliberate rather than
incidental.

DanCardin:
I dont think we benefit much, if at all, by testing pytest vs pytest-asyncio vs python versions matrix-wise. Either these dependent libraries work universally across the combinations or they dont, but I dont think there's a meaningful way (unless shown otherwise) that those combinations really become pytest-alembic specific.

Rather, we **do** benefit from making sure that pytest-alembic is compatible with each unique depdendency/python version individually.

Fixes https://github.com/schireson/pytest-alembic/issues/176

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,15 +48,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
-        pytest-version: ["7.0.0", "8.3.5"]
-        pytest-asyncio-version: ["0.16.0", "0.23.0"]
-        sqlalchemy-version: ["2.0.40"]
+        include:
+          # Sweep across python version sweep with the highest checked dependency version.
+          # Older pytest/asyncio versions are tested separately below on a single
+          # interpreter since we have no reason to believe interpreter-specific breakage
+          # would surface in those older versions but not in pytest 9 + asyncio 1.x,
+          # but we **do** know that newer versions of dependencies may not support arbitrarily old
+          # python versions.
+          - {python-version: "3.10", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.11", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.12", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.13", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.14", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.15", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
 
-        exclude:
-          # old versions of pytest-asyncio use old pytest hook syntax
-          - pytest-version: "8.3.5"
-            pytest-asyncio-version: "0.16.0"
+          # For older tool versions, we can just stick to a single python version, again, because we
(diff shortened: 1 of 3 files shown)


tschm:
testing against 3.15 is very ambitious

tschm:
I think we should accept pytest > 9 in order to be able to support 3.14.

tschm:
@DanCardin I recommend to close this one

DanCardin:
We can do revisions to the matrix after the fact. I just wanted to prove 3.14 (and 3.15) work first, to roughly equivalent coverage to what we had before

Issue #203, opened 2026-08-26
main carries an unreleased breaking change under the previous version


tschm:
Sub-issue of #196 (quality assessment of `main` @ `d62ef3a`).

**Subcategory:** public API / semver discipline — **8 → 10**

## Why

`pyproject.toml:3` still reads `version = "0.12.1"`, which is also the newest tag
(`v0.12.1`). Since that tag, `194ecb0` (`chore!: drop Python 3.9, raising the floor to
3.10`) raised `requires-python` to `>=3.10` — a breaking change, correctly marked `!` per
the convention convco reads.

Nothing is wrong *yet*: the release has not happened. The risk is that the version string
and the `!` commit live in different places and nothing in CI connects them, so the next
release can be cut as `0.12.2` without anything objecting. For a `0.x` package the
breaking bump is the minor: **`0.13.0`**.

## done when

Either the version is bumped to `0.13.0` on `main`, or `release.yml` gains a check that
the tag being released is consistent with the conventional commits since the previous
one.

## Evidence

`git tag --sort=-v:refname | head -1` → `v0.12.1`; `pyproject.toml:3` →
`version = "0.12.1"`; `194ecb0` is a `chore!:` commit after that tag.

tschm:
Re-checked during a `/rhiza:quality` run on `main` @ `da41949`. **The condition this
issue describes is still present**, so flagging it rather than leaving it closed
silently — PR #206 was closed unmerged and nothing else appears to have taken its place.

The published release and `main` disagree on the Python floor:

```
$ curl -s https://pypi.org/pypi/pytest-alembic/json | jq -r '.info | "\(.version)  \(.requires_python)"'
0.12.1  <4,>=3.9
```

```
$ grep -n 'requires-python\|^version' pyproject.toml
5:version = "0.12.1"
18:requires-python = ">=3.10,<4"
```

So `main` drops Python 3.9 while still carrying the version number of a release that
declares `>=3.9`. Anyone installing from git, or building a sdist from this tree, gets
something materially different from PyPI's `0.12.1` under the same name and version.

No action taken on my side — reopening or re-closing with a rationale is a maintainer
call. If the intent is that 3.9 support ends with 0.13.0 (#190), then this resolves
itself at the version bump and the issue is fine to leave closed; it just isn't resolved
on `main` today.

PR #216, opened 2026-08-26
build: declare and gate the dependency floors, requiring pytest 9 and sqlalchemy 2

### LINKED ISSUE #211: Remove the unused requires_sqlalchemy_2 marker, or restore the sqlalchemy axis it needs
**Subcategory:** test design quality — 8 → 9

`requires_sqlalchemy_2` is defined at `tests/__init__.py:12` and referenced nowhere:

```
$ grep -rn requires_sqlalchemy_2 tests src
tests/__init__.py:12:requires_sqlalchemy_2 = pytest.mark.skipif(not is_sqlalchemy_2, reason="Requires v2")
```

`is_sqlalchemy_2` (`:6`) exists only to build it, so both lines are dead. They are
counted as covered because importing the module executes them — a case where the 100%
floor reports on code no test consults.

The marker is also unreachable by construction. `.github/workflows/build.yml:53` pins
`sqlalchemy-version: ["2.0.40"]`, so `is_sqlalchemy_2` is `True` in every CI entry and
the skip can never fire. `pyproject.toml`'s coverage comment already records this:

> The previous floor of 93 was set because the CI matrix skipped the asyncio tests on
> sqlalchemy versions without asyncio support. That reason is gone: the matrix pins a
> single sqlalchemy (2.0.40), so no entry skips them any more.

**Scope note:** the sibling symbols are *not* dead. `requires_asyncio_support` is applied
at `tests/test_runner.py:264`, `:270` and `:280`, and `sqlalchemy_version` /
`supports_asyncio` feed it. Only the `requires_sqlalchemy_2` / `is_sqlalchemy_2` pair
should be removed.

**done when** `tests/__init__.py` defines no unused symbol — either the pair is deleted,
or a sqlalchemy axis is restored to the matrix and the marker is applied to the tests
that need it.

### LINKED ISSUE #212: Declare version floors for alembic and sqlalchemy
**Subcategory:** dependency & security hygiene — 9 → 10

`pyproject.toml:19` declares two of the three runtime dependencies with no version
specifier at all:

```toml
dependencies = ["pytest>=7.0", "alembic", "sqlalchemy"]
```

Confirmed in the built wheel's metadata (`uv build`, then reading
`pytest_alembic-0.12.1.dist-info/METADATA`):

```
Requires-Dist: pytest>=7.0
Requires-Dist: alembic
Requires-Dist: sqlalchemy
```

`pytest` carries a floor; `alembic` and `sqlalchemy` do not. So a resolver is free to
install sqlalchemy 1.3 alongside this package, and the failure surfaces at import or
first use rather than at resolve time — where it would be actionable.

The repo already knows which versions it supports:

- `.github/workflows/build.yml:54` tests `sqlalchemy-version: ["2.0.40"]` only.
- `src/pytest_alembic/executor.py:22` and `runner.py:20` import
  `sqlalchemy.engine.Connectable`, and `executor.py:331` imports
  `sqlalchemy.ext.asyncio.AsyncEngine`, which needs 1.4+.
- `tests/__init__.py` still carries 1.3/1.4 feature detection that CI cannot exercise
  (see the sibling finding on `requires_sqlalchemy_2`).

Settling on the floor is the substantive part of this: the declared minimum should be a
version the project is willing to claim, not just the one CI happens to pin.

**done when** `alembic` and `sqlalchemy` carry lower bounds in `pyproject.toml`, the
bounds match what the code actually requires, and `make build` produces a wheel whose
`Requires-Dist` lines show them.

tschm:
Closes #212.

Three commits, each independently droppable. The first is #212 as filed; the second and
third came out of running the mechanism that verifies it.

## 1. `build:` declare version floors for alembic and sqlalchemy

`alembic` and `sqlalchemy` were declared bare, and had been since the initial commit --
`alembic = "*"` / `sqlalchemy = "*"` under poetry, carried over verbatim by the uv
migration. So a resolver was free to pair this package with a sqlalchemy it cannot run
on, and the failure surfaced at import rather than at install.

The floors come from running the suite, not from the CI pin. The check that matters:

```
$ uv pip install pytest_alembic-0.12.1-py3-none-any.whl 'sqlalchemy==1.3.24'
× No solution found when resolving dependencies:
  ╰─▶ ... pytest-alembic==0.12.1 depends on sqlalchemy>=1.4 ...
```

A resolver error instead of a runtime `ModuleNotFoundError`, which is the whole point.

## 2. `ci:` gate the floors with a lowest-direct resolution job

A declared floor that nothing installs is a claim, not a constraint. This adds
`lowest-deps`, which resolves every direct dependency to its declared minimum
(`uv sync --all-groups --resolution lowest-direct`) and runs the suite there. The
mechanism is taken from [jebel-quant/rhiza](https://github.com/jebel-quant/rhiza), which
runs the same job in its own CI.

It paid for itself immediately, finding three wrong floors including one from commit 1:

- **`alembic>=1.7` was wrong** and is raised to `>=1.9`. 1.7.0 passes against sqlalchemy
  1.4 -- which is how I verified it -- and fails at import against sqlalchemy 2 with
  `NameError: name 'TextClause' is not defined` in `alembic/operations/ops.py`. Alembic
  does not upper-bound sqlalchemy, so nothing else in the resolver rejects that pair.
  Nothing else in this workflow *builds* that pair either: the matrix installs one
  sqlalchemy over an ordinary sync, so the old-alembic/new-sqlalchemy corner was
  unreachable until this job existed.
- **`greenlet>=3.0`** → `>=3.1`. 3.0.0 predates Python 3.13, ships no wheel for it, and
  the sync fell back to compiling from source and failed.
- **`sphinx-autobuild` unbounded** resolved to v0.1.0, from 2013, which builds only
  against Python's long-removed `imp` module. The other bare entries in `dev` and `docs`
  get floors for the same reason.

## 3. `build!:` require pytest 9 and sqlalchemy 2

**Breaking.** Raises two floors to what this repository actually exercises, and deletes
the code that existed only for what is below them.

`pytest>=9`: this is a pytest plugin, so pytest is its host rather than an ordinary
dependency -- loaded by a pytest too old for it, the failure lands inside that pytest's
startup, for every project that merely has this package installed. The floor also aligns
with `requires-python`, since pytest 9 itself requires `>=3.10`.

`sqlalchemy>=2.0`: dropping 1.x is a support decision, not a technical one -- the suite
passes on 1.4.24 and 1.4.54. What 1.x does not have is verification. Nothing in this
repository tests it: the matrix installs a 2.0.x for every entry, the dev group needs
`sqlalchemy[asyncio]>=2.0.42`, the 42 example projects pin nothing, and `lowest-deps`
resolves to the dev group's higher floor. A `1.4` bound would be claiming support that
nothing checks -- the exact failure mode this branch removes.

Deleted as a direct consequence, all dead at these floors: the `DeclarativeMeta` and
`AsyncEngine` import fallbacks in `executor.py` and the two experimental tests (both
symbols resolve unconditionally on 2.0, verified even with greenlet absent); the
sqlalchemy version guards in `tests/__init__.py` (**closes #211**); the
`pytest_asyncio.fixture` AttributeError fallback in the async example's conftest; and
`anyio` from the dev group, which nothing imports -- `deptry` could not see it, since it
scans `src` only.

The matrix moves with the floors: `pytest-version` → 9.0.0/9.1.1 and
`pytest-asyncio-version` → 1.0.0/1.4.0, floor and current on both axes. The old
7.0.0/8.3.5 and 0.16.0/0.23.0 entries tested combinations the package no longer claims,
and the `exclude` pairing old pytest-asyncio with new pytest goes with them.

## Verification

| Gate | Result |
| --- | --- |
| `make lint` | pass -- ruff check, ruff format, mypy over 34 modules |
| `make deps` | pass |
| `make docs-coverage` | pass -- 120/120, floor 100 |
| `make audit` / `make audit-all` | pass -- lockfile in sync, `--frozen` export resolves |
| `make docs` | pass -- sphinx `-W`, zero warnings |
| `make test` | **150 passed**, coverage `882 stmts / 196 branches, 0 miss, 100%` |
| `lowest-deps` (new) | **149 passed, 1 skipped** at pytest 9.0.0, pytest-asyncio 1.3.0, alembic 1.9.0, sqlalchemy 2.0.42 |

The one skip is `test_pyproject_toml`, which `_supports_toml()` correctly reports as
unsupported on alembic 1.9. That is also why `lowest-deps` runs bare `pytest` rather than
`make test`: `make test` owns the 100% coverage gate, which one skipped test cannot meet.

## Two things deliberately left alone

- **No version bump.** This is breaking, but `main` already carries the unreleased Python
  3.9 drop under 0.12.1 (#203 -- closed, though the condition is still observable: PyPI's
  0.12.1 declares `>=3.9`, `main` declares `>=3.10`, and #206 was closed unmerged). The
  release decision belongs in one place rather than in this PR.
- **The matrix's `~=` is not a pin.** `sqlalchemy~=2.0.40` means `>=2.0.40,<2.1`, so CI
  installs **2.0.52** today and silently follows every new 2.0.x. The comment in
  `pyproject.toml` describing "the matrix pins a single sqlalchemy (2.0.40)" is therefore
  describing a pin that does not exist. Worth a separate decision -- correct the comment,
  or make it a real `==`.


tschm:
Flagging an overlap before this goes further: **#215 revises the same matrix block in the
same file**, and the two PRs disagree about something substantive rather than merely
conflicting textually.

This PR removes the `pytest 7.0.0 / 8.3.5` and `pytest-asyncio 0.16.0 / 0.23.0` axes on
the grounds that they test combinations below the declared `pytest>=9` floor. #215 does
close to the opposite: it restructures the matrix into `include:` form and *keeps* those
combinations deliberately, as single-interpreter entries anchored on 3.13, with comments
explaining why each pairing exists.

#215's own CI run is the part worth weighing, because it is evidence rather than opinion.
On `main`, at 14:16 today:

```
test (3.13, 7.0.0, 0.16.0, 2.0.40)   success
test (3.13, 8.3.5, 0.23.0, 2.0.40)   success
test (3.14, 9.1.1, 1.4.0, 2.0.40)    success
test (3.15, 9.1.1, 1.4.0, 2.0.40)    failure   <- the only failure in that run
```

So pytest 7.0.0 and 8.3.5 demonstrably still work with this package, and someone is
actively investing in keeping them tested. That is a direct argument against the
`pytest>=9` floor in commit 3 here: unlike the sqlalchemy 1.x case in the same commit --
where nothing in the repository tested 1.x, which was the whole justification for
dropping it -- pytest 7/8 *are* tested and *do* pass. The symmetry I claimed in that
commit message does not hold on the pytest side.

Reasonable resolutions, in rough order of how much of this PR survives:

1. **Drop the pytest floor to `>=7.0` and revert the matrix hunk**, keeping the sqlalchemy
   2.0 floor, the alembic 1.9 floor and the `lowest-deps` job. #215 then owns the matrix
   outright and there is no conflict. Note the `lowest-deps` job would need the dev group
   to keep a modern pytest, since `pytest-asyncio 1.x` is what pytest 9 requires.
2. **Keep `pytest>=9`** and treat #215's old-pytest entries as testing an unsupported
   configuration -- which is coherent only if dropping pytest 7/8 support is actually
   intended, and that is a maintainer call, not something a floor should decide by
   accident.
3. **Merge #215 first** and rebase this PR onto it, resolving the matrix in whichever
   direction (1) or (2) settles.

Also worth noting for #176: #215 shows **Python 3.14 passing**, and 3.15 as the only red
cell in that run.

Separately, CI has not run on this PR at all. GitHub Actions has been in a critical
incident since 15:11:58Z today ("we've throttled inbound traffic", 15:48) and this branch
was pushed at 15:49:45Z, so the trigger was dropped -- zero runs and zero check-runs on
both the fork and here. The workflow file itself is fine: `actionlint` 1.7.12 exits clean,
and `on.push.branches: ["**"]` is intact at the pushed ref. I will re-trigger once Actions
is operational; the local gate results in the PR description stand in the meantime.

tschm:
**Correcting my earlier comment on this PR.** I claimed #215's CI showed pytest 7.0.0 and
8.3.5 passing, and used that to argue the `pytest>=9` floor here should be reverted. That
was wrong. Those jobs did not run pytest 7 or 8. From the log of #215's
`test (3.13, 7.0.0, 0.16.0, 2.0.40)`:

```
```

**The matrix version axes are inert**, and have been since they were introduced. Every
`make` target invokes its tool through `uv run`, which re-syncs the environment to
`uv.lock` before running, undoing the `uv pip install` pins from the preceding step. The
same log shows the whole sequence:

```
uv sync                    ->  + pytest==9.1.1
uv pip install pytest~=7.0 ->  - pytest==9.1.1
                               + pytest==7.0.1
make test (uv run ...)     ->  platform linux -- ... pytest-9.1.1
```

So sixteen matrix entries have been testing one combination while reporting sixteen
different ones in their names. `pytest-version`, `pytest-asyncio-version` and
`sqlalchemy-version` have never affected what ran. That also means there is no evidence
either way about pytest 7/8 support -- my argument against the floor evaporates, and so
does any claim that the old entries were protecting something.

The latest commit here fixes that with `UV_NO_SYNC: "1"` on the linter and test steps,
and fixes the two real bugs that inertness had been hiding:

- **pytest 9.0.0 is not usable with this package.** It calls argparse's
  `parse_known_intermixed_args` with a namespace already carrying `file_or_dir`; some
  CPython 3.12 patch releases warn about that, and `filterwarnings = ["error"]` promotes
  the warning to an INTERNAL_ERROR inside the `pytester` sub-runs, so `test_disabled_cli`
  gets exit code 3 instead of 5. Measured: on CPython 3.12.3, 9.0.0 fails and 9.0.1,
  9.0.2, 9.0.3, 9.1.0, 9.1.1 all pass; 3.12.7 fails likewise, 3.12.13 does not. Floor
  raised to `pytest>=9.0.1`.
- **pytest-asyncio 1.0 and 1.2 declare `pytest<9`**, so the `pytest-asyncio>=1.0` floor
  and the `1.0.0` matrix entry named a combination that cannot resolve at all. 1.3.0 is
  the first release admitting pytest 9 -- which is what `--resolution lowest-direct` had
  been picking anyway. Both now say 1.3.

Verified locally with the pins genuinely applied, across the four combinations the matrix
now names: (9.0.1, 1.3.0), (9.0.1, 1.4.0), (9.1.1, 1.3.0), (9.1.1, 1.4.0) -- 150 passed
each.

One consequence worth flagging for #215: with `UV_NO_SYNC` in place the matrix becomes
real, so any old-pytest entries it keeps will genuinely install old pytest for the first
time. Whether pytest 7/8 still work here is now an open question rather than a settled
one, and it is answerable by running that matrix rather than by reading its previous
green checkmarks.

tschm:
@DanCardin careful here. We say sqlalchemy >= 2.0 and pytest >= 9.0 This does simplify a few things or are we going too far?

pyproject.toml
```diff
@@ -14,7 +14,58 @@ classifiers = [
 ]
 readme = "README.md"
 requires-python = ">=3.10,<4"
-dependencies = ["pytest>=7.0", "alembic", "sqlalchemy"]
+# Floors, not pins -- the point is that an unusable combination fails at resolve
```
DanCardin, line 17: i dont want all of this reasoning in the pyproject.toml

tschm: You need lower bounds for all dependencies. Otherwise you have no chance to make sure a feasible combination will work

tschm: You could of course remove comments from pyproject

DanCardin (commented):
I *know* the package works with older pytest versions, and older pytest-asyncio versions. I also know that for decent span of pytest-asyncio versions, they had horrible backwards incompatible issues that were broadly broken, so i broadly dont know what's going on in >1.0

We lost sqlalcehmy 1.4 testing at some point, but i **think** we should also generally be broadly compatible with it as well. i dont necessarily feel strongly about support of it if it turns out i'm wrong and it's *not* supported in the current version already, but if it does work today, then i'd rather add back its version to be tested.

I also have an open issue for pytest-anyio, so I dont particularly want pytest-asyncio as a required first-party dependency here either. Ostensibly, this library **should** be agnostic to the asyncio adapter

pyproject.toml
DanCardin: i'm willing to add lower bounds, but im not particularly interested in substantially increasing the lower bounds beyond what the code is physically incompatible with right now.

tschm: I think it boils down to whether you want Python 3.14 or support older versions of sqlalchemy and packages. But I will have a closer look again

tschm: For me, versions of this project are made for exactly that purpose. If you want to use Python 3.10 and sqlalchemy 1.4 then use the existing version but going forward only sqlalchemy 2 is supported

DanCardin: I just dont see much value in dropping support. Dropping python 3.7/8 had a meaningful benefit in code style/quality. sqlalchemy 1.4 does not take any additional effort to support at this time. 1.4 was also getting support until ~2 years ago. 

pytest, i'm **maybe** more ambiguous about, I'd be more apt to drop 7 because it removes the conditional method defs from plugin.py, and that release is like 3 years ago.

pytest-asyncio, i think we may internally even still have consumers of 0.19 era versions, which are barely a year old. Again, I'd be more apt to stop testing older versions necessarily, but this package doesn't **actually** depend upon it at all.

PR #219, opened 2026-08-26
chore: drop the pytest-asyncio test dependency

tschm:
`pytest-alembic` itself never used `pytest-asyncio`. The shipped package depends on `pytest`, `alembic` and `sqlalchemy`, and drives async engines with a bare `asyncio.run` in `executor.run_task`; nothing under `src` imports the plugin. It was a dev dependency only.

## Three of its four usages were dead weight

- `examples/test_async_sqlalchemy` decorated a **synchronous** fixture with `pytest_asyncio.fixture`, and set `asyncio_mode = auto` in its `setup.cfg` for it. Nothing there was async.
- The root config set `asyncio_default_fixture_loop_scope`, and then added a `filterwarnings` ignore for the very deprecation warning that setting exists to silence.
- `anyio` sat in the dev group entirely unused.

## The one real usage did not need a plugin either

Two examples built `alembic_engine` from `create_postgres_fixture(async_=True)`, which is a genuine async fixture and so needs a plugin to resolve. But `pytest-alembic` never awaits your fixture — it only requires that one *yields* an `AsyncEngine`. A plain `@pytest.fixture` works, and then no plugin is needed at all.

So the two examples now cover the two synchronous forms `docs/source/asyncio.rst` already documents, one each. That also makes their test docstrings accurate for the first time:

| example | form | docstring claim |
|---|---|---|
| `test_async_sqlalchemy` | `AsyncEngine(create_engine(url))` | "manually adapted" ✔ |
| `test_async_sqlalchemy_native` | `create_async_engine(url)` | "native" ✔ |

Previously both built their engines natively, despite one claiming to cover the manually adapted case.

Since the pmr `async_=True` form is no longer exercised here, `asyncio.rst` now says plainly that it needs an async-fixture plugin in the reader's own project, and that the two plain-fixture forms need none. That requirement belongs to the fixture, not to `pytest-alembic`, and was previously unstated.

## CI

The `pytest-asyncio-version` dimension (`0.16.0` / `0.23.0` / `1.4.0`) is gone. It was never testing `pytest-alembic` against `pytest-asyncio` — no code path touches it — it was testing `pytest-mock-resources` against `pytest-asyncio`. The `pytest` 7.0.0 and 8.3.5 entries survive, now with a comment recording why those two: 7.0.0 is the floor declared in `[project.dependencies]`, and 8.3.5 covers the intervening major whose hook API differs from both neighbours.

`contributing.rst` also claimed CI exercises 3.10 through 3.13; the matrix is 3.10 through 3.15. Corrected in the same sentence.

## Verification

With both packages absent from the venv:

- `150 passed`
- coverage **100%**, so the floor in `[tool.coverage.report]` still holds
- `ruff check` / `ruff format --check` / `mypy` / `deptry` / `interrogate` clean
- `sphinx-build -W` succeeds
- the three async examples pass under `pytest~=7.0.0` and `pytest~=8.3.5`, the two older CI pins

Runtime is unchanged — 27.89s mean over three full-suite runs, against 28.00s before.

## A note on uv.lock

The lock diff is a pure 27-line deletion, hand-edited. A plain `uv lock` (uv 0.11.28) rewrites roughly 190 lines of unrelated resolution markers, and `uv lock --check` passes on the committed lock, so that rewrite is normalization rather than anything required. Worth a glance in case CI's uv version disagrees.

Also worth recording for anyone editing a lock by hand: `uv lock --check` did **not** catch a first attempt that accidentally stripped `anyio` from `starlette` and `watchfiles`' dependency lists. That was found by diffing the dependency graph, and the edit redone scoped to the `pytest-alembic` package block.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,26 +50,24 @@ jobs:
       matrix:
         include:
           # Sweep across python version sweep with the highest checked dependency version.
-          # Older pytest/asyncio versions are tested separately below on a single
-          # interpreter since we have no reason to believe interpreter-specific breakage
-          # would surface in those older versions but not in pytest 9 + asyncio 1.x,
+          # Older pytest versions are tested separately below on a single interpreter
+          # since we have no reason to believe interpreter-specific breakage would
+          # surface in those older versions but not in pytest 9,
           # but we **do** know that newer versions of dependencies may not support arbitrarily old
           # python versions.
-          - {python-version: "3.10", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
-          - {python-version: "3.11", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
-          - {python-version: "3.12", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
-          - {python-version: "3.13", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
-          - {python-version: "3.14", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
-          - {python-version: "3.15", pytest-version: "9.1.1", pytest-asyncio-version: "1.4.0", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.10", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.11", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.12", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.13", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.14", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
+          - {python-version: "3.15", pytest-version: "9.1.1", sqlalchemy-version: "2.0.40"}
 
           # For older tool versions, we can just stick to a single python version, again, because we
           # have no reason to believe their pytest-alembic specifically impacts their compatibility.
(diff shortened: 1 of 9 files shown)


tschm:
CI is green across all 14 checks, and the test banner confirms the plugin is genuinely gone rather than merely uninstalled from the manifest:

```
plugins: alembic-0.12.1, mock-resources-2.13.0
```

No `asyncio-`, no `anyio-`. Every job reports that.

## One caveat on the older-pytest jobs, which is pre-existing and not introduced here

The job named `test (3.13, 7.0.0, 2.0.40)` did **not** run pytest 7.0.0. Its banner says:

```
```

The `uv pip install 'pytest~=7.0.0'` step does take effect, but the next step (`make lint`) uses a plain `uv run`, which re-syncs against the lockfile and reinstalls pytest 9.1.1 — visible as `Uninstalled 1 package` under **Run linters** — before `make test` ever starts. So both older-pytest jobs have been exercising 9.1.1 for as long as the pattern has been in place. The same applies to the `sqlalchemy~=` pin.

This predates this PR and I have deliberately left it alone, since fixing it means changing `make lint` / `make test` to `uv run --no-sync` (or `--frozen`), which affects every job rather than just the async ones. Happy to do it as a follow-up if you want.

It does mean the "passes under 7.0.0 and 8.3.5" claim in the description rests on local verification rather than on these jobs. That verification is real — I re-ran it with the banner visible to be sure the pin actually held:

```
platform darwin -- Python 3.13.14, pytest-7.0.1, pluggy-1.6.0
plugins: alembic-0.12.1, mock-resources-2.13.0
3 passed, 38 deselected

platform darwin -- Python 3.13.14, pytest-8.3.5, pluggy-1.6.0
plugins: alembic-0.12.1, mock-resources-2.13.0
3 passed, 38 deselected
```

Worth noting the fix would make those two jobs meaningfully stricter than they are today, so it is possible they surface something once they genuinely run 7.0.0.

DanCardin (approved):
hmm. I really really thought I recalled that fixture annotation being required, interesting.

Given that anyio is known non-functional, i'm expecting anyio to become a dev requirement again anyways but that's neither here nor there.

I also sort of expect to regain pytest-asyncio tbh. it seems like a lack of coverage that we have no tests of manually written pytest tests that happen to be async.

PR #221, opened 2026-08-27
Declare dependency floors, gate them in CI, and fix async commit on sqlalchemy 1.4

### LINKED ISSUE #212: Declare version floors for alembic and sqlalchemy

tschm:
## Summary

`alembic` and `sqlalchemy` have been declared bare — no version bound at all — since the initial commit, so `pip install pytest-alembic` was free to resolve a combination this package cannot run on. This PR gives them floors, adds a CI job that proves the floors are real, and fixes the one genuine incompatibility that job uncovered.

## The gate

A `lowest-deps` job runs the suite under `uv sync --resolution lowest-direct`, which pins every *direct* dependency to the minimum its specifier allows. Without it, a floor is just a number somebody typed; with it, a wrong floor fails CI.

`lowest-direct` rather than `lowest`: transitive dependencies stay at normal versions. Flooring those too tests other people's release histories rather than this project's declared support, and fails on packages that never shipped a wheel for a supported Python.

It paid for itself immediately, catching four separate problems:

| Declared | Problem |
|---|---|
| `alembic>=1.7` | passes against sqlalchemy 1.4, fails at import against sqlalchemy 2 (`NameError: name 'TextClause' is not defined`) |
| `greenlet>=3.0` | predates Python 3.13, no wheel for it |
| `sphinx-autobuild` (bare) | resolved to v0.1.0 from 2013, which builds only against Python's removed `imp` |
| `sqlalchemy[asyncio]>=2.0.42` in the dev group | masked a real async bug — see below |

## The bug it found

The dev group carried `sqlalchemy[asyncio]>=2.0.42`. Since the higher floor wins, `lowest-deps` resolved sqlalchemy to 2.0.42 and never exercised the 1.4 end that `dependencies` claims to support. Removing that override dropped the resolution to 1.4.24 and the async tests failed at once:

```
AttributeError: 'Connection' object has no attribute 'commit'
  .../sqlalchemy/ext/asyncio/engine.py:288
```

`ConnectionExecutor.run_task` drove async engines with `engine.connect()` plus an explicit `await connection.commit()`. `AsyncConnection.commit` does not exist before sqlalchemy 2.0, so **every** 1.4 release raised this for any user passing an async engine — verified across 1.4.24 (the lowest the suite can reach), 1.4.31, 1.4.40, 1.4.46 and 1.4.54 (the last). The declared `sqlalchemy>=1.4` bound was true only for sync engines.

The dev pin was hiding a real incompatibility rather than recording one, so the fix went into `executor.py`, not back into the pin:

```python
async with engine.begin() as connection:
    result = await connection.run_sync(fn, **kwargs)
```

`begin()` is equivalent on the happy path — it commits on success and rolls back on an exception — and exists across the entire declared range. Full suite results after the change:

| sqlalchemy | result |
|---|---|
| 1.4.24 | 149 passed, 1 skipped |
| 1.4.54 | 149 passed, 1 skipped |
| 2.0.0 | 149 passed, 1 skipped |
| 2.0.52 | 150 passed |

## The floors

`sqlalchemy>=1.4` — `executor.py` imports `sqlalchemy.ext.asyncio`, which does not exist before 1.4.0; on 1.3.24 the suite does not even collect. The bound stays at 1.4 rather than 1.4.24 (the lowest the *suite* can reach, since pytest-mock-resources excludes 1.4.0–1.4.23): that is a development constraint, and imposing it on an installing user would refuse a combination that works for them.

`alembic>=1.9` — a compatibility boundary, not an API one. Every alembic symbol this package touches resolves as far back as 1.0.11. What fixes the number is sqlalchemy 2 support, which alembic gained in 1.9. Notably this is *not* about pyproject.toml support: `_supports_toml()` feature-detects `Config.get_alembic_option`, and `test_legacy_alembic_without_toml_support` covers the fallback.

The dev/docs groups get floors for the same reason, plus one specific to them — under `lowest-direct`, an unbounded entry does not mean "any version is fine", it means the oldest release on PyPI.

Closes #212 — `alembic` and `sqlalchemy` now carry lower bounds, the bounds match what the code actually requires, and the built wheel's metadata shows them:

```
Requires-Dist: pytest>=7.0
Requires-Dist: alembic>=1.9
Requires-Dist: sqlalchemy>=1.4
Requires-Python: >=3.10, <4
```

#211 stays open — the `requires_sqlalchemy_2` marker is untouched here. The sqlalchemy-axis question it raises is discussed [in that thread](https://github.com/schireson/pytest-alembic/issues/211#issuecomment-5434418167).

## Verification

All gates green locally: `make lint` (ruff + mypy), `make deps` (deptry), `make test` (150 passed, **100% coverage**), and the exact `lowest-deps` command at the floors (alembic 1.9.0, sqlalchemy 1.4.24, pytest 7.0.0, greenlet 3.1.0) — 149 passed, 1 skipped.


tschm:
@DanCardin Careful! A small correction in executor.py is introduced. I glad you insisted on using lower bounds. It works now.

.github/workflows/build.yml
```diff
@@ -193,6 +193,51 @@ jobs:
...
+    # than this project's declared support, and fails on packages that never published a
+    # wheel for a supported Python.
+    #
+    # It covers `sqlalchemy>=1.4` too, which it briefly did not: the dev group used to
+    # carry `sqlalchemy[asyncio]>=2.0.42`, and since the higher floor wins, the
+    # resolution landed on 2.0.42 and never exercised the 1.4 end at all. Removing that
+    # override dropped the resolution to 1.4.24 and immediately failed `run_task`, which
+    # called `AsyncConnection.commit()` -- a method that does not exist before
+    # sqlalchemy 2.0. The dev pin had been hiding a real incompatibility rather than
+    # recording one, so the fix went into `executor.py`, not back into the pin.
```
```suggestion
```

.github/workflows/build.yml
```diff
@@ -193,6 +193,51 @@ jobs:
...
+          cache-dependency-glob: "uv.lock"
+
+      - name: Install the project at its declared floors
+        run: uv sync --all-groups --resolution lowest-direct
+
+      - name: Run the suite at the declared floors
+        # Not `make test`: that runs under coverage against the 100% floor, which the
+        # oldest supported alembic cannot meet -- `test_pyproject_toml` skips there,
+        # since `_supports_toml()` correctly reports no pyproject.toml support. This job
+        # asks whether the code runs at its floors, and `make test` already owns coverage.
```
```suggestion
        # Not `make test`: that runs under coverage against the 100% floor.
```

src/pytest_alembic/executor.py
```diff
@@ -337,9 +337,17 @@ def run_task(self, fn: Callable[..., Any], **kwargs: Any) -> Any:
...
-                async with engine.connect() as connection:
+                # `begin()`, not `connect()` followed by an explicit `commit()`. The two
+                # are equivalent on the happy path -- both commit on success and roll
+                # back on an exception -- but `AsyncConnection.commit` does not exist
+                # before sqlalchemy 2.0, so the explicit form raises `AttributeError:
+                # 'Connection' object has no attribute 'commit'` against every 1.4
+                # release, from 1.4.24 (the lowest the suite can reach) through 1.4.54
+                # (the last). `begin()` is available across the whole declared range,
+                # which is what makes the `sqlalchemy>=1.4` floor in pyproject.toml
+                # true for async engines and not only for sync ones.
```
DanCardin:
```suggestion
```

pyproject.toml
```diff
@@ -15,7 +15,37 @@ classifiers = [
...
+# the suite passes on alembic 1.7.0 against sqlalchemy 1.4. What fixes the number is
+# sqlalchemy 2: alembic gained 2.0 support in 1.9, and 1.7.0 paired with sqlalchemy
+# 2.0.42 fails at import with `NameError: name 'TextClause' is not defined` from
+# `alembic/operations/ops.py`. Alembic does not upper-bound sqlalchemy, so nothing else
+# rejects that pair -- this floor is what does.
+#
+# Newer alembic is not required for the pyproject.toml support: `_supports_toml()`
+# feature-detects `Config.get_alembic_option`, and
+# `test_legacy_alembic_without_toml_support` covers the fallback path, which is why the
+# floor stays at 1.9 rather than following alembic's latest.
```
```suggestion
```

pyproject.toml
```diff
@@ -15,7 +15,37 @@ classifiers = [
...
+dependencies = ["pytest>=7.0", "alembic>=1.9", "sqlalchemy>=1.4"]
```
DanCardin, line 18: I dont really consider the alembic/sqlalchemy intercompatibility my problem, someone using alembic and sqlalchemy would have hit that before installing this plugin. but i also dont think the alembic constraint is as problematic so meh

pyproject.toml
```diff
@@ -28,20 +58,30 @@ requires = ["uv_build>=0.6,<0.13"]
 build-backend = "uv_build"
 
 [dependency-groups]
+# Floored for the same reason the runtime dependencies above are, plus one specific to
+# these groups: the `lowest-deps` CI job resolves every *direct* dependency to its
+# declared minimum (`uv sync --resolution lowest-direct`), so an unbounded entry here
+# does not mean "any version is fine" -- it means that job installs the oldest release
+# on PyPI. Left bare, `sphinx-autobuild` resolves to v0.1.0 (2013), which builds only
+# against Python's long-removed `imp` module and fails the sync outright.
```
```suggestion
```

DanCardin (approved):

pyproject.toml
tschm: well, alembic is today at version 1.19. So the discussion whether the lower bound should be 1.7 or 1.9 is weird. The test introduced here would not pass with 1.7, so 1.9 makes sense

PR #237, opened 2026-08-27
docs: expand the README badge row

tschm:
Three badges become twelve, taking the layout from
[jebel-quant/jquantstats](https://github.com/jebel-quant/jquantstats). Grouped into three
rows so the block reads as three statements rather than one wall:

| row | badges |
|---|---|
| **the package** | PyPI version · Python versions · licence · downloads/month |
| **what CI proves** | CI · coverage · docs · OpenSSF scorecard |
| **the toolchain** | ruff · mypy · uv · pre-commit |

## Every URL was fetched first, and checked for what it *renders*

A 200 is not enough — shields will happily serve a badge reading `invalid`. Actual
rendered text:

| badge | renders |
|---|---|
| PyPI version | `pypi \| v0.12.1` |
| Python versions | `Python \| 3.10 • 3.11 • 3.12 • 3.13 • 3.14 • 3.15` |
| downloads | `PyPI downloads per month \| 1M` |
| OpenSSF scorecard | `openssf scorecard \| 7` |
| ruff / mypy / uv / pre-commit | as expected |

## Two of jquantstats' badges are deliberately not carried over

- **Rhiza** — this repo is not rhiza-managed, so the badge would have nothing to read.
- **CodeFactor** — `codefactor.io/repository/github/schireson/pytest-alembic/badge`
  returns **404**; the project isn't registered there. A badge that renders an error is
  worse than an absent one.

## The Python row is hardcoded on purpose

The obvious choice, `img.shields.io/pypi/pyversions/pytest-alembic`, reads the
**published** classifiers. Today that renders:

```
python | 3.9 | 3.10 | 3.11
```

— advertising a version that has been dropped, and omitting four the matrix gates. So it
is hardcoded, with a comment above the block recording that it tracks
`.github/workflows/build.yml` and `[project.classifiers]`.

**Coupling worth noting:** this lists 3.15, matching the CI matrix, same as #234 does for
the classifiers. 3.15 is currently `3.15.0rc1` — if you'd rather not advertise an rc, drop
`%20%E2%80%A2%203.15` here and the `3.15` line in #234 together, and the two stay in step.

## Verification

- All URLs absolute — `docs/source/quickstart.rst` includes this file verbatim into sphinx,
  where a repo-relative link would not resolve, so the licence badge points at the GitHub
  blob rather than at `LICENSE`
- `make docs` — succeeds under `-W` from a clean tree; all **12** images confirmed present
  in the generated `quickstart.html`
- Example checker — README's 3 fences still parse, 0 violations
- `pre-commit run --all-files` — passes

The existing CI badge also picks up the link it was missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,25 @@
-![Github Actions Build](https://github.com/schireson/pytest-alembic/actions/workflows/build.yml/badge.svg)
+<!-- Badges are grouped: the package itself, then what CI proves about it, then the
+     toolchain. Every URL is absolute -- this file is included verbatim into the sphinx
+     docs by docs/source/quickstart.rst, where a repo-relative link would not resolve.
+     The Python row is hardcoded rather than using img.shields.io/pypi/pyversions,
+     which reads the *published* classifiers and so currently renders "3.9 | 3.10 |
+     3.11" -- including a version that has been dropped. Keep it in step with the
+     matrix in .github/workflows/build.yml and the classifiers in pyproject.toml. -->
+
+[![PyPI version](https://img.shields.io/pypi/v/pytest-alembic)](https://pypi.org/project/pytest-alembic/)
+[![Python versions](https://img.shields.io/badge/Python-3.10%20%E2%80%A2%203.11%20%E2%80%A2%203.12%20%E2%80%A2%203.13%20%E2%80%A2%203.14%20%E2%80%A2%203.15-blue?logo=python&logoColor=white)](https://www.python.org/)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](https://github.com/schireson/pytest-alembic/blob/main/LICENSE)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,25 @@
-![Github Actions Build](https://github.com/schireson/pytest-alembic/actions/workflows/build.yml/badge.svg)
+<!-- Badges are grouped: the package itself, then what CI proves about it, then the
+     toolchain. Every URL is absolute -- this file is included verbatim into the sphinx
+     docs by docs/source/quickstart.rst, where a repo-relative link would not resolve.
+     The Python row is hardcoded rather than using img.shields.io/pypi/pyversions,
+     which reads the *published* classifiers and so currently renders "3.9 | 3.10 |
+     3.11" -- including a version that has been dropped. Keep it in step with the
+     matrix in .github/workflows/build.yml and the classifiers in pyproject.toml. -->
+
```
DanCardin:
```suggestion
```

Issue #241, opened 2026-08-27
Quality assessment of main @ b336cd5 — 9.6/10


tschm:
Quality assessment of `main` at `b336cd5`, run with `/rhiza:quality`.

**Supersedes #223** (assessed at `7638580`, 9.3/10). The score moves **9.3 → 9.6**.
The subcategory set is identical to #223's — same 17, same names — so the two numbers are
directly comparable; per-subcategory movement is in the table below.

**All eight findings from #223 were closed, and seven are verified fixed here.** That is
the whole of the movement: **+6 across five subcategories**, against **three newly
measured findings** that hold three subcategories at 9 rather than moving them to 10.
Nothing #196, #214 or #223 investigated and withdrew is re-derived — the two long-standing
retracted arguments were re-tested this run and stay retracted (see *What stays
withdrawn*).

**The one finding that closed without a fix is #232**, closed by me with no comment and no
remediation. Its condition is still observable at `b336cd5`, so it is carried forward
below rather than quietly dropped — see finding 4.

## How this was measured, and what it does not cover

This repo is **not rhiza-managed** — there is no `.rhiza/` directory (all three step-0
probes negative) — so the assessment ran in **degraded mode**:

- **No Rhiza gate ran.** `fmt`, `typecheck`, `security` and `rhiza-test` are not defined
  in this profile. The marks rest on this repo's **own** discovered targets plus a
  hand-gathered design analysis.
- **Template fidelity is not applicable** — there is no template to be faithful to.
- Because there is no template, the `Makefile` and `.github/workflows/` are this repo's
  own work and are **in scope**. That is why *CI wiring*, *toolchain reproducibility* and
  *contributor onboarding* appear below; they have no counterpart in a managed repo's
  scorecard, so don't read a managed repo as having scored 10 on them.
- **The result rests on a narrower base than a full run and is not comparable to one.**

Language profile: Python, source root `src`, manifest `pyproject.toml`, lockfile
`uv.lock`, toolchain pin `.python-version` (3.13).

## Gate results

| Gate | How it resolved | Result |
| --- | --- | --- |
| fmt + typecheck | `make lint` — ruff check, ruff format --check, `mypy src tests` | **PASS** — clean on 279 files / 34 modules |
| docs-coverage | `make docs-coverage` (rung 1) | **PASS** — interrogate 120/120, floor 100 |
| deps | `make deps` (rung 1) | **PASS** — no dependency issues, 17 files |
[...]

PR #253, opened 2026-08-28
docs: record why MigrationContext is one class, not two

### LINKED ISSUE #249: `runner.py` is both the largest module and the coupling hotspot
Finding 6 of #241 (assessment of `main` @ `b336cd5`, 9.6/10). *Overall architecture, 9 → 10.*

The weakest of the six findings, and included only because the mark is below 10.

`src/pytest_alembic/runner.py` is 521 lines, the largest module in `src` by 117 lines
(`plugin/plugin.py` is 404). It is imported by five of the seventeen modules —
`__init__.py`, `tests/default.py`, both experimental test modules, and `plugin/fixtures.py` —
the highest in-degree in the graph. Its `MigrationContext` carries both the migration verbs
(`migrate_up_one`, `migrate_down_to`, `roundtrip_next_revision`) and the history/state
accessors (`current`, `heads`, `table_at_revision`, `insert_into`), and holds the two
worst-but-still-`B` blocks in the package's non-test code (`managed_downgrade` `B (8)` at
`:261`, `managed_upgrade` `B (6)` at `:238`).

Nothing here is a defect: MI is `A (54.45)`, no block reaches `C`, and `MigrationContext` is
the documented public surface — so splitting it is an API decision, not a refactor. The other
half of the mark, the `plugin.plugin` → `tests` → `plugin.error` cycle, is deliberate and
correctly explained at `plugin/plugin.py:257-262`, and should stay.

**done when** either the two responsibilities in `MigrationContext` are separated behind the
existing public names, or a note records that its size is a deliberate consequence of being
the public API. A mark of 10 here is not obviously worth an API change — the note is the
recommended path, and it is self-certification rather than a fix, which is worth saying out
loud.

tschm:
Closes #249.

`runner.py` is the largest module in the package (521 lines, 117 more than `plugin/plugin.py`)
and has the highest in-degree in the import graph — five of seventeen modules import it. And
`MigrationContext` carries both the migration verbs (`migrate_up_one`, `migrate_down_to`,
`roundtrip_next_revision`) and the history/state accessors (`current`, `heads`,
`table_at_revision`, `insert_into`).

Read as a class in isolation, that is two responsibilities and an obvious split. Which is
why it keeps being raised — three assessments in a row have now named it, most recently as
finding 6 of #241.

It is one class because it is *the* documented public surface: the object `alembic_runner`
yields, and the receiver of every call in every documented example. Splitting it renames or
re-homes names users have written tests against, for no defect — no block exceeds `B`,
maintainability index is `A (54.45)`. So this records the decision instead of making the
change, which is the second branch of the issue's `done when`, and points a future split at
delegation behind the existing public names.

A comment rather than a docstring addition, deliberately: the audience is the next person to
propose the split, not the user reading the rendered API page.

No behaviour change. `make lint` clean.

### DIFF
--- a/src/pytest_alembic/runner.py
+++ b/src/pytest_alembic/runner.py
@@ -65,6 +65,20 @@ def runner(config: "Config", engine: Connectable | None = None) -> Iterator["Mig
     yield migration_context
 
 
+# On this module's size, which is deliberate. `runner.py` is the largest module in the
+# package and the most imported one, and `MigrationContext` below carries both the
+# migration verbs (`migrate_up_one`, `migrate_down_to`, `roundtrip_next_revision`) and the
+# history/state accessors (`current`, `heads`, `table_at_revision`, `insert_into`). Read as
+# a class in isolation, that is two responsibilities and an obvious split.
+#
+# It is one class because it is *the* documented public surface: the object
+# `alembic_runner` yields, and the receiver of every call in every example in the docs.
+# Splitting it would rename or re-home names that users have written tests against, for no
+# defect -- the metrics are healthy (no block above `B`, maintainability index `A`), so the
+# split would buy a tidier graph at the cost of an API break.
+#
+# If a genuine reason to split arrives, keep the public names on this class and delegate,
+# rather than asking callers to reach for a second object.
 @dataclass
 class MigrationContext:
     """Within a given environment/execution context, executes alembic commands.
--- a/src/pytest_alembic/runner.py
+++ b/src/pytest_alembic/runner.py
@@ -65,20 +65,9 @@ def runner(config: "Config", engine: Connectable | None = None) -> Iterator["Mig
     yield migration_context
 
 
-# On this module's size, which is deliberate. `runner.py` is the largest module in the
-# package and the most imported one, and `MigrationContext` below carries both the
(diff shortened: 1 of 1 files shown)


src/pytest_alembic/runner.py
```diff
@@ -65,6 +65,20 @@ def runner(config: "Config", engine: Connectable | None = None) -> Iterator["Mig
...
+# a class in isolation, that is two responsibilities and an obvious split.
+#
+# It is one class because it is *the* documented public surface: the object
+# `alembic_runner` yields, and the receiver of every call in every example in the docs.
+# Splitting it would rename or re-home names that users have written tests against, for no
+# defect -- the metrics are healthy (no block above `B`, maintainability index `A`), so the
+# split would buy a tidier graph at the cost of an API break.
+#
+# If a genuine reason to split arrives, keep the public names on this class and delegate,
+# rather than asking callers to reach for a second object.
```
DanCardin:
```suggestion
# This class may seem to be overly large and have a, perhaps, over broad set of responsibilities.
# However it represents the public API of a migration, containing all the utilities a migration test
# author (users of this library) should require to manipulate migrations within their tests.
```

Issue #250, opened 2026-08-28
CI installs `tomli`/`coveralls` outside the lockfile, and the scanning workflows are not in the tree


tschm:
Carried from #241's *toolchain reproducibility* row (currently 10), where both items are
noted and **deliberately not scored**. Filed so they are tracked rather than re-noted a
third time — and because they are the likeliest candidates for a future assessment to measure
at greater depth and dock a mark that is 10 today.

### 1. `tomli` and `coveralls` are installed unpinned and outside the lockfile

`.github/workflows/build.yml`, in two places:

- the `Coveralls` step of every `test` matrix row: `uv pip install tomli coveralls`
- the `finish` job: `pip install tomli coveralls`

Neither is in `pyproject.toml` or `uv.lock`, so a coveralls release can change what CI runs
without a commit to this repo, and `finish` resolves against whatever `pip` finds. Every
*other* tool CI uses is either locked or a pinned `uvx` download.

Fix: declare them in a dependency group (a `ci` group, kept out of `dev`) and install through
`uv sync --group ci`, or failing that pin both with `==`.

### 2. The CodeQL and Dependency-Graph workflows are not in `.github/workflows/`

Both run on every push, so they are enabled through GitHub's default setup rather than
checked in. The consequence is narrow but real: that configuration cannot be reviewed in a
pull request, does not travel with a fork or a clone, and cannot be rolled back with a
revert.

Fix: commit them as workflow files. Worth noting for whoever picks this up — GitHub's
code-scanning *default setup* has to be switched off before an advanced/committed CodeQL
workflow will run, so this needs repository-settings access and is not a PR on its own.

**done when** the CI-only packages resolve from the lockfile or a pinned specifier, and the
scanning configuration is reviewable in the tree (or a note records that default setup is
deliberate).

PR #254, opened 2026-08-28
ci: resolve coveralls and tomli from the lockfile instead of installing them ad hoc

tschm:
Addresses the first half of #250. The second half — committing the CodeQL and
Dependency-Graph workflows — needs code-scanning *default setup* switched off in repository
settings first, so it is not part of this PR.

Two steps installed CI-only packages from outside the lockfile:

- every `test` matrix row: `uv pip install tomli coveralls`
- the `finish` job: a bare `pip install tomli coveralls`, in a job with no checkout and no uv
  at all

Every other tool CI uses is either locked or a SHA-pinned action, so these two were the one
place a third-party release could change what CI ran with no commit to this repo. #241
carried this as a hygiene note under *toolchain reproducibility* twice without scoring it.

### What changes

`coveralls` and `tomli` now live in a `ci` dependency group. That puts them in `uv.lock`,
inside `make audit-all`'s `--all-groups` sweep, and in dependabot's view.

The `test` row syncs them with `uv sync --frozen --inexact --only-group ci`. Both flags
matter: `--only-group` and `--inexact` keep the sync from disturbing the sqlalchemy and
pytest pins the earlier steps installed. The re-sync trap #239 fixed for `uv run` applies
just as well to a careless `uv sync` here, and this step runs after the tests but its output
is still attributed to the matrix cell.

The `finish` job gains the checkout and setup-uv steps it needs for the lockfile to be
present at all, and runs `uv run --no-sync coveralls --service=github --finish`.

`make deps` gains `--group ci`, for the reason it already passes `--group docs`: so deptry
introspects the real package instead of printing an "Assuming the corresponding module name
is ..." note on every run.

### Verified

- the sync form was run locally: `uv sync --frozen --inexact --only-group ci` adds the group
  without removing anything, and `coveralls --help` resolves afterwards
- `coveralls` 4.1.0's CLI still takes `--service` and `--finish`, so neither invocation
  changes
- the `uv.lock` diff is additions only — `coveralls` plus `typer`, `rich`, `shellingham`,
  `annotated-doc`; no version changes to existing packages and no removals
- `make audit-all` no known vulnerabilities across all groups; `make deps` clean and now
  silent; `make lint` clean; `make test` 150 passed at 100%

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,11 +110,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
         run: |
-          # From the lockfile, through the `ci` group. An ad-hoc `uv pip install coveralls`
-          # let a coveralls release change what this step ran with no commit to this repo.
-          # `--only-group --inexact` so the sync adds the group without disturbing the
-          # version pins the earlier steps installed, and `uv run` does not re-sync
-          # because UV_NO_SYNC is set at the job level.
+          # Install the CI deps without disturbing the
+          # version pins from earlier steps due to UV_NO_SYNC.
           uv sync --frozen --inexact --only-group ci
           uv run coveralls --service=github
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,7 +110,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
         run: |
-          uv pip install tomli coveralls
+          # From the lockfile, through the `ci` group. An ad-hoc `uv pip install coveralls`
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -110,7 +110,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
         run: |
-          uv pip install tomli coveralls
+          # From the lockfile, through the `ci` group. An ad-hoc `uv pip install coveralls`
+          # let a coveralls release change what this step ran with no commit to this repo.
+          # `--only-group --inexact` so the sync adds the group without disturbing the
+          # version pins the earlier steps installed, and `uv run` does not re-sync
+          # because UV_NO_SYNC is set at the job level.
```
```suggestion
          # Install the CI deps without disturbing the
          # version pins from earlier steps due to UV_NO_SYNC.
```

PR #255, opened 2026-08-28
docs: set nitpicky so an unresolved cross-reference fails the docs gate

### LINKED ISSUE #248: Set `nitpicky` so an unresolved cross-reference fails the docs gate, and state the supported platforms
Finding 5 of #241 (assessment of `main` @ `b336cd5`, 9.6/10). *User-facing documentation, 9 → 10.*

`docs/source/conf.py` sets no `nitpicky` (`grep -c nitpicky` → `0`). #227 is fixed and
verified in the built HTML, but the reason `-W` could not catch it is unchanged: with
`nitpicky` unset, an unresolved Python cross-reference is **silently dropped** rather than
warned about, so it never reaches `-W`. The next stale `intersphinx_mapping` entry, or the
next renamed sqlalchemy symbol, degrades the rendered page and passes the docs gate.

### Verified before filing

Trial build at `1db9f8b` from a cleared `docs/build/`, `sphinx-build -n -b html`: **17
warnings, 10 unique** — the whole backlog, and all of it small.

Adding `"alembic": ("https://alembic.sqlalchemy.org/en/latest/", None)` to
`intersphinx_mapping` clears 3 of them (`alembic.config.Config` ×3) — verified, the inventory
resolves. The remaining 14 are internal and each is a one-line fix:

| Count | Unresolved target | Shape of fix |
| ---: | --- | --- |
| 4 | `:func:`dict`` ×3, `:func:`list`` | `:class:` or a literal |
| 7 | `:func:`alembic_runner`` ×3, `runner` ×2, `create_alembic_fixture`, `alembic_config` | fully-qualified target, or autodoc the fixtures |
| 3 | `MigrationContext`, `pytest_alembic.MigrationContext`, `RevisionSpec` | paths autodoc does not publish |

Second item under the same mark, carried from #223 finding 7's second half: with
`cross-platform` now running Windows and macOS, no document states which platforms are
supported. There is no `Operating System` classifier at all
(`grep -n "Operating System" pyproject.toml` → no match), and neither the README nor the docs
mention an OS.

### Proposed change

- add the alembic intersphinx mapping, fix the 14 internal references, then set
  `nitpicky = True` with `nitpick_ignore` for anything deliberately unresolvable
- add `Operating System :: OS Independent` to the classifiers and one README line naming
  Linux / macOS / Windows, matching what `cross-platform` gates

**done when** `nitpicky = True` is set so an unresolved reference fails the `-W` build, and
the supported platforms are stated in the README or the classifiers.

tschm:
Closes #248.

`make docs` already builds under `-W`, but with `nitpicky` unset sphinx **silently drops**
an unresolved Python cross-reference and renders the text unlinked — there is nothing for
`-W` to see. That is exactly how the stale intersphinx entry fixed in #227 got past the docs
gate: the references pointed at SQLAlchemy 1.3 targets that no longer existed, the pages
lost their links, and the build stayed green.

Setting `nitpicky = True` surfaced the whole standing backlog — **17 warnings, 10 unique**.
All 17 are fixed here, and **no `nitpick_ignore` was needed**.

| Fixed | How |
| --- | --- |
| `alembic.config.Config` ×3 | `intersphinx_mapping` gains alembic's inventory |
| `alembic_runner` ×3, `alembic_config`, `create_alembic_fixture` | qualified: the fixtures live in `plugin.fixtures`, so a bare `:func:`alembic_runner`` never resolved outside that module's own autodoc context |
| `dict` ×3, `list` | they are `:class:`, not `:func:` |
| `MigrationContext` ×2, `pytest_alembic.MigrationContext` | referenced as `pytest_alembic.runner.MigrationContext`, the path autodoc publishes |
| `runner` ×2 | `api.rst` now documents `pytest_alembic.runner.runner` alongside `MigrationContext`; it is public and exported, and was already referenced from a docstring that could not resolve it |
| `RevisionSpec` | qualified to `pytest_alembic.revision_data.RevisionSpec` |

### Verified in the built HTML, not just by the absence of warnings

- `api.html` → `https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config`
- `custom_data.html` → `https://docs.python.org/3/library/stdtypes.html#dict` and `#list`
- `custom_tests.html` → `api.html#pytest_alembic.runner.MigrationContext`

`make docs` succeeds under `-W` with `nitpicky` on, from a cleared `docs/build/`.

### Second half of the issue: the supported platforms are now stated

The `cross-platform` job added in #240 gates macOS and Windows, but nothing said so — the
classifiers named interpreters only, and there was no `Operating System` classifier at all.
This adds `Operating System :: OS Independent` and a README line giving the Python, alembic
and SQLAlchemy ranges together with what CI runs where.

`make lint` clean, `make test` 150 passed at 100%.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ See the full documentation [here](https://pytest-alembic.readthedocs.io/en/lates
 A pytest plugin to test alembic migrations (with default tests) and
 which enables you to write tests specific to your migrations.
 
+Runs on Python 3.10 – 3.15 with alembic 1.9+ and SQLAlchemy 1.4+, on Linux, macOS and
+Windows. CI runs the whole suite on Linux, and on macOS and Windows every test that does
+not need a database server.
+
 ```bash
 $ pip install pytest-alembic
 $ pytest --test-alembic
--- a/README.md
+++ b/README.md
@@ -21,8 +21,7 @@ A pytest plugin to test alembic migrations (with default tests) and
 which enables you to write tests specific to your migrations.
 
 Runs on Python 3.10 – 3.15 with alembic 1.9+ and SQLAlchemy 1.4+, on Linux, macOS and
-Windows. CI runs the whole suite on Linux, and on macOS and Windows every test that does
-not need a database server.
+Windows.
 
 ```bash
 $ pip install pytest-alembic
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,15 +45,12 @@ html_sidebars = {"**": ["globaltoc.html", "relations.html", "sourcelink.html", "
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "sqlalchemy": ("https://docs.sqlalchemy.org/en/20/", None),
-    # alembic's own inventory, so `:class:`alembic.config.Config`` resolves to alembic's
-    # page rather than being dropped.
     "alembic": ("https://alembic.sqlalchemy.org/en/latest/", None),
 }
 
 # Every unresolved cross-reference is a warning, and `make docs` builds under `-W`, so it
 # is an error. Without this, sphinx silently drops an unresolved Python reference and
-# renders the text unlinked -- which is how the stale intersphinx entry fixed in #227 got
-# past a `-W` build in the first place: there was nothing for `-W` to see.
+# renders the text unlinked.
 nitpicky = True
 
 autoclass_content = "both"
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,7 +45,16 @@ html_sidebars = {"**": ["globaltoc.html", "relations.html", "sourcelink.html", "
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "sqlalchemy": ("https://docs.sqlalchemy.org/en/20/", None),
+    # alembic's own inventory, so `:class:`alembic.config.Config`` resolves to alembic's
+    # page rather than being dropped.
+    "alembic": ("https://alembic.sqlalchemy.org/en/latest/", None),
 }
+
+# Every unresolved cross-reference is a warning, and `make docs` builds under `-W`, so it
+# is an error. Without this, sphinx silently drops an unresolved Python reference and
+# renders the text unlinked -- which is how the stale intersphinx entry fixed in #227 got
+# past a `-W` build in the first place: there was nothing for `-W` to see.
+nitpicky = True
 
 autoclass_content = "both"
 master_doc = "index"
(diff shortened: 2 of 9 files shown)


docs/source/conf.py
```diff
@@ -45,7 +45,16 @@
...
     "sqlalchemy": ("https://docs.sqlalchemy.org/en/20/", None),
+    # alembic's own inventory, so `:class:`alembic.config.Config`` resolves to alembic's
+    # page rather than being dropped.
+    "alembic": ("https://alembic.sqlalchemy.org/en/latest/", None),
 }
 
+# Every unresolved cross-reference is a warning, and `make docs` builds under `-W`, so it
+# is an error. Without this, sphinx silently drops an unresolved Python reference and
+# renders the text unlinked -- which is how the stale intersphinx entry fixed in #227 got
+# past a `-W` build in the first place: there was nothing for `-W` to see.
```
DanCardin:
```suggestion
# Every unresolved cross-reference is a warning, and `make docs` builds under `-W`, so it
# is an error. Without this, sphinx silently drops an unresolved Python reference and
# renders the text unlinked.
```

docs/source/conf.py
```diff
@@ -45,7 +45,16 @@
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "sqlalchemy": ("https://docs.sqlalchemy.org/en/20/", None),
+    # alembic's own inventory, so `:class:`alembic.config.Config`` resolves to alembic's
+    # page rather than being dropped.
```
DanCardin:
```suggestion
```

pyproject.toml
```diff
@@ -8,6 +8,9 @@ license-files = ["LICENSE"]
 keywords = ["pytest", "sqlalchemy", "alembic", "migration", "revision"]
 classifiers = [
     "Framework :: Pytest",
+    # The cross-platform CI job runs the database-free tests on Windows and macOS, and the
+    # matrix runs everything on Linux, so this is gated rather than assumed.
```
```suggestion
```

README.md
```diff
@@ -20,6 +20,10 @@ See the full documentation [here](https://pytest-alembic.readthedocs.io/en/lates
 A pytest plugin to test alembic migrations (with default tests) and
 which enables you to write tests specific to your migrations.
 
+Runs on Python 3.10 – 3.15 with alembic 1.9+ and SQLAlchemy 1.4+, on Linux, macOS and
+Windows. CI runs the whole suite on Linux, and on macOS and Windows every test that does
+not need a database server.
```
DanCardin:
```suggestion
Windows.
```

PR #256, opened 2026-08-28
test: gate undefined names inside doctest examples, `>>> def` bodies included

### LINKED ISSUE #246: Nothing checks names inside `>>> def` doctest bodies
Finding 3 of #241 (assessment of `main` @ `b336cd5`, 9.6/10). *Executable documentation, 9 → 10.*

This is the mechanism gap named when closing #230, filed here rather than left in a closed
thread.

`create_alembic_fixture`'s example bound the fixture to `alembic` and then referenced
`alembic_runner` in the body — an undefined name that would raise `NameError` for anyone who
copied it. #236 fixed the instance. It had survived a 100%-coverage suite,
`--doctest-modules`, and an example checker's `--run` pass, because a `>>> def …: …` body is
*parsed* and never *executed*: doctest sees a function definition and never attempts name
resolution inside it.

26 of the 60 examples under `src` are of exactly this shape — they need a live fixture and a
database, so they legitimately cannot carry doctest output. That is not the gap. The gap is
that no tool reads inside them.

### Verified before filing

1. **No off-the-shelf gate exists.** ruff 0.16.4 does not lint code inside docstrings. Given
   a probe containing precisely the #236 bug shape:

   ```python
   def f():
       """Doc.

       >>> def test_x(alembic):
       ...     alembic_runner.migrate_up_one()
       """
   ```

   `ruff check --isolated --select F821` reports *All checks passed*.

2. **A ~30-line checker does work.** Walk the package, `doctest.DocTestFinder().find(module)`,
   concatenate each `DocTest`'s `ex.source`, `ast.parse`, run `pyflakes.checker.Checker`, and
   report only `UndefinedName`.

3. **One design detail decides whether it works at all.** Seeding the checker's `builtins`
   with the module's globals is the obvious choice, and it silently masks the very bug this
   is for — `alembic_runner` *is* a module global in `plugin/fixtures.py`. Measured both ways
   over `src/pytest_alembic` at `1db9f8b`:

   | Seeding | Result |
   | --- | --- |
   | module globals (`vars(mod)`) | 28 doctests, **0** violations — the `alembic_runner` shape would be invisible |
   | the documented object's own qualname components only | 28 doctests, **3** violations, all real |

4. **The 3 residual violations are genuine.** All in `alembic_config`'s docstring
   (`src/pytest_alembic/plugin/fixtures.py:124-133`): the examples use `@pytest.fixture` and
   `alembic.config.Config()` with no `>>> import pytest` / `>>> import alembic.config` line,
   so nobody can copy-run them.

### Proposed change

- add `pyflakes` to the dev group and the checker as `make doctest-names` plus a CI step (or
  as `tests/test_doctest_names.py`, so it rides the existing suite)
- seed only the documented object's own name, not the module namespace, and say why in a
  comment — that is the difference between a gate and a no-op
- add the missing `>>> import` lines to `alembic_config`'s two examples

**done when** an undefined name inside a `>>> def` body fails a gate — locally or in CI —
rather than shipping.

tschm:
Closes #246.

The example fixed in #236 bound the fixture to `alembic` and then called `alembic_runner` in
the body — an undefined name that raised `NameError` for anyone who copied it. It got past a
100%-coverage suite, `--doctest-modules`, and a docs build, because a `>>> def ...` body is
only *compiled*: doctest sees a function definition and never resolves the names inside it.
26 of the 60 examples under `src` are that shape, since they need a live fixture and a
database and so cannot carry doctest output. The instance was fixed; nothing stopped the
next one.

ruff does not lint code inside docstrings — checked against exactly that shape with
`--select F821`, which reports *All checks passed* — so this reads the examples out with
`doctest.DocTestFinder` and runs pyflakes over them, one test case per documented object.
It rides the existing suite rather than adding a CI job, so it runs in every matrix cell and
fails where a contributor is already looking.

### The seeding is the part that matters

An example may reference the object being documented and, for a method, its class. Anything
else it must import, because an example is something a reader copies into their own file.

The wider and more obvious choice — seed `vars(module)` — **makes this gate a no-op**:
`alembic_runner` is itself a module-level name in `plugin.fixtures`, so seeding the module
namespace finds zero problems in this package and would have let #236's bug through
unchanged. Measured both ways:

| Seeding | Result |
| --- | --- |
| `vars(module)` | 0 violations — the bug this exists for is invisible |
| documented object + its class | 3 violations, all real |

`test_the_check_sees_inside_a_def_body` pins that down against the historical bug's shape, so
a later "simplification" of the seeding fails loudly rather than silently disarming the check.

### The three violations it found are real

All in `alembic_config`'s docstring: the examples used `@pytest.fixture` and
`alembic.config.Config()` with no import line, so neither could be copied and run. Fixed by
importing what they use — which also documents what to import.

### Gates

- `make lint` clean (mypy included; the new module is annotated)
- `make test` 179 passed (150 + 28 documented objects + the check's own regression test), coverage unchanged at `885 stmts / 196 branches, 0 miss, 100%`
- `make deps`, `make docs` clean
- the floors job simulated locally, since this adds a dev dependency:
  `uv sync --all-groups --resolution lowest-direct` then
  `uv run --all-groups --resolution lowest-direct pytest src tests` → 178 passed, 1 skipped
  (the known `test_pyproject_toml` skip on alembic 1.9.0)

The `uv.lock` diff adds `pyflakes` only — re-locked with `--no-upgrade` so no other package
moves. The remaining churn in that file is uv rewriting dependency markers, the same
rewrite `f9bdc24` produced.

### Confirmed running in CI, not just locally

Read out of the job logs rather than inferred from the green tick: the parametrised cases and
`test_the_check_sees_inside_a_def_body` appear in **every** job — all eight `test` rows,
`lowest-deps`, and `cross-platform` on both Windows and macOS — at 179 passed per row.

### DIFF
--- a/tests/test_doctest_examples.py
+++ b/tests/test_doctest_examples.py
@@ -1,0 +1,95 @@
+"""Undefined names in doctest examples, as a gate -- including inside `>>> def` bodies.
+
+`--doctest-modules` executes the examples under ``src``, but 26 of them consist of a bare
+``def test_...(alembic):`` whose body calls into the fixture. Those need a live fixture and
+a database, so they cannot carry doctest output and are not meant to run: doctest only
+*compiles* the function definition and never resolves the names inside its body.
+
+That is how the example fixed in #236 shipped. It bound the fixture to one name and then
+called another in the body -- an undefined name that raised `NameError` for anyone who
+copied it -- and got past a 100%-coverage suite, `--doctest-modules`, and a docs build.
+
+(No example in this docstring, deliberately: `--doctest-modules` collects this file too, and
+an illustration of the broken shape would be collected as a doctest of its own.)
+
+ruff does not lint code inside docstrings (checked against exactly that shape with
+`--select F821`), so this reads the examples out with `doctest` and runs pyflakes over them.
+"""
+
+import ast
+import doctest
+import importlib
+import pkgutil
+from types import ModuleType
+
+import pytest
+from pyflakes.checker import Checker
+from pyflakes.messages import UndefinedName
+
+import pytest_alembic
+
+
+def _modules() -> list[ModuleType]:
+    """Import every module in the package, so `DocTestFinder` can be given the objects."""
+    modules = [pytest_alembic]
+    for module in pkgutil.walk_packages(pytest_alembic.__path__, f"{pytest_alembic.__name__}."):
+        modules.append(importlib.import_module(module.name))
+    return modules
+
+
+def _doctests() -> list[doctest.DocTest]:
+    finder = doctest.DocTestFinder(exclude_empty=True)
+    return [test for module in _modules() for test in finder.find(module) if test.examples]
+
+
+def _permitted_names(name: str) -> set[str]:
+    """The names an example may use without importing them.
+
+    Deliberately narrow: the object being documented, and its class if it is a method. An
+    example is something a reader copies into their own file, so anything else it references
+    it should import -- and the wider, more obvious choice of seeding the module's whole
+    namespace makes this gate a no-op. `alembic_runner` *is* a module-level name in
+    `plugin.fixtures`, so seeding `vars(module)` finds zero problems in this package and
+    would have let #236's bug through unchanged.
+    """
+    parts = name.split(".")
+    return set(parts[-2:])
+
+
+def _undefined_names(test: doctest.DocTest) -> list[str]:
+    """Report the undefined names across one docstring's examples."""
+    # Examples in one docstring share a namespace, in doctest and for a reader working
+    # through them in order, so they are checked as one unit rather than one at a time.
+    source = "".join(example.source for example in test.examples)
+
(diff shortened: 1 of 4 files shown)


tests/test_doctest_examples.py
```diff
@@ -0,0 +1,95 @@
...
+``def test_...(alembic):`` whose body calls into the fixture. Those need a live fixture and
+a database, so they cannot carry doctest output and are not meant to run: doctest only
+*compiles* the function definition and never resolves the names inside its body.
+
+That is how the example fixed in #236 shipped. It bound the fixture to one name and then
+called another in the body -- an undefined name that raised `NameError` for anyone who
+copied it -- and got past a 100%-coverage suite, `--doctest-modules`, and a docs build.
+
+(No example in this docstring, deliberately: `--doctest-modules` collects this file too, and
+an illustration of the broken shape would be collected as a doctest of its own.)
```
DanCardin:
```suggestion
```

